PR #31, opened 2026-05-21
Fix privacy toggle scoping

### LINKED ISSUE #30: Unauthenticated /privacy/toggle lets any client disable processing for the entire instance
Why this is major
The privacy flag is process-wide global state in [privacy.py:1], and the audio pipeline checks that global flag before doing any work in [pipeline.py:21]. Because the route is exposed without any auth or per-user scoping in [privacy.py:13] and is registered publicly in [main.py:179], any unauthenticated caller can flip the system into private mode and cause all audio ingestion to be skipped for every user.

Issue-ready description
The [/privacy] endpoints are public and mutate shared application state. [POST /privacy/toggle] changes a module-level [PRIVATE_MODE] flag, and [process_audio()] checks that flag globally before transcription or storage. This means one request from any client can turn off memory processing across the whole deployment, effectively creating an unauthenticated denial of service for the core pipeline.

Steps to reproduce

Send [POST /privacy/toggle] with no credentials.
Confirm the response returns the new private state.
Submit a normal audio record or pipeline request afterward.
Observe that processing is skipped because [process_audio()] exits early when [is_private()] is True.
Expected behavior
Privacy controls should be authenticated and scoped to the correct user or tenant. One client should not be able to change ingestion behavior for everyone.

Actual behavior
Any unauthenticated client can toggle the shared private mode and disable processing for all users until it is toggled back.

Suggested fix
Require authentication on the privacy endpoints and persist privacy state per user or tenant instead of using a process-global variable. If this is intended to be an admin-only switch, gate it behind an admin role or service secret. Add tests covering unauthorized access and ensuring privacy changes do not affect other users.

Aryan-B-Parikh:
## Summary
- Require authenticated access for privacy endpoints
- Scope privacy state per user instead of using a shared process-wide flag
- Update the audio pipeline to respect the authenticated user’s privacy state
- Add tests covering unauthorized access and user isolation

## Root cause
The privacy toggle was public and controlled a global in-memory flag, so any client could disable processing for all users.

## Changes
- Added JWT-based auth dependency to GET /privacy and POST /privacy/toggle
- Replaced global privacy state with per-user state
- Updated process_audio to check privacy for the current user
- Added endpoint and helper tests

## Validation
- Python syntax checks passed for all edited files
- Full pytest run could not be executed locally because required dependencies are not installed in this environment

##Closes : #30

### DIFF
--- a/backend/app/routes/privacy.py
+++ b/backend/app/routes/privacy.py
@@ -1,15 +1,16 @@
-from fastapi import APIRouter
+from fastapi import APIRouter, Depends
 
+from app.services.auth import get_current_user_id
 from app.services.privacy import is_private, toggle_privacy
 
-router = APIRouter()
+router = APIRouter(prefix="/privacy", tags=["privacy"])
 
 
 @router.get("/")
-def get_privacy():
-    return {"private": is_private()}
+def get_privacy(user_id: str = Depends(get_current_user_id)):
+    return {"private": is_private(user_id)}
 
 
 @router.post("/toggle")
-def toggle():
-    return {"private": toggle_privacy()}
+def toggle(user_id: str = Depends(get_current_user_id)):
+    return {"private": toggle_privacy(user_id)}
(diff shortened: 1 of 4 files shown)


Chetan0e:
@Aryan-B-Parikh merged , thank you for contribution

PR #34, opened 2026-05-22
fix: resolve SECRET_KEY security misconfiguration in .env.example

### LINKED ISSUE #23: SECRET_KEY placeholder in .env.example is misleading and a security risk
The .env.example file contains a malformed SECRET_KEY placeholder that poses a subtle but serious security risk for new contributors and anyone following the setup guide.
Current value: SECRET_KEY=generate-a-long-random-secret-key-herettings

This appears to be a leftover from an incomplete edit — the intended instruction text (generate-a-long-random-secret-key-here) was merged with what looks like a fragment of the word "settings", producing a value that reads like a placeholder but is actually 39 characters of seemingly random text.
Why this is dangerous
The backend enforces a minimum SECRET_KEY length of 32 characters. This placeholder is 39 characters long, which means it silently passes validation. A developer who:

clones the repo
runs cp .env.example .env
skips carefully reading every line
starts the server

...will have a running application signing JWTs with a publicly known secret key. There is no warning, no startup error, and nothing in the README that flags this specific value as dangerous. Every token issued by that instance is trivially forgeable by anyone who has read the repo.
This is qualitatively different from a placeholder like your-secret-here or an empty value — those would either fail validation or obviously not work. This one works, which is the problem.

Steps to reproduce

cp .env.example .env
Fill in only MONGO_URI, GROQ_API_KEY, and GEMINI_API_KEY (the only values the README calls out as required)
Start the backend with python run.py
Observe: server starts successfully with no warning about the default SECRET_KEY
Any attacker who has read .env.example can now forge valid JWT tokens for any user

Expected behavior
The SECRET_KEY placeholder should either:

Be an empty value (SECRET_KEY=) so the server refuses to start, or
Be an obviously invalid string (e.g. SECRET_KEY=REPLACE_THIS_BEFORE_RUNNING) that fails the 32-character check, or
Trigger a startup check that compares the live value against known insecure defaults and exits with a clear error message

Affects: .env.example, backend/app/config.py, README.md
Severity: High — silent security misconfiguration, no user-visible warning

Vishakhayadavv:
Fixes #23

Changes limited to:
- `.env.example` — set `SECRET_KEY=` (empty) so the server refuses to start if not configured
- `README.md` — replaced the bad placeholder with a generation command

No unrelated changes included.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,7 +27,7 @@ MONGO_URI=mongodb+srv://your-username:your-password@cluster0.d7dlkdt.mongodb.net
 DATABASE_NAME=verath
 
 # Security
-SECRET_KEY=generate-a-long-random-secret-key-herettings
+SECRET_KEY=
 
 # Audio Settings
 AUDIO_SAMPLE_RATE=16000
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
         cd backend
         pip install -r requirements.txt
         pip install pytest pytest-asyncio pytest-cov httpx
+        pip install ruff
     
     - name: Run linter
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,8 @@ jobs:
       with:
         python-version: '3.11'
         cache: 'pip'
+    - name: Install system dependencies
+      run: sudo apt-get install -y portaudio19-dev
     
     - name: Install dependencies
       run: |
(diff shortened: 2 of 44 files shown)


Chetan0e:
@Vishakhayadavv 
Thanks for the contribution.

I’m closing this PR for now because it includes many unrelated commits and changes beyond the original issue scope. Some of the commits are already being handled under other issues/PRs, and I do not want to cherry-pick commits for a small scoped fix.

Please pull the latest project changes first, create a fresh branch, and make a clean PR containing only the necessary changes related to this issue.

PR #40, opened 2026-05-22
perf: push export filters to MongoDB and stream JSON response

KaparthyReddy:
- Add filtered_memories() to memory_store.py: builds MongoDB query with intent and date range filters at DB level using metadata.intent and created_at $gte/$lte — only matching documents are transferred
- Replace all_memories(limit=10000) + Python-side filtering in export endpoint with filtered_memories() — eliminates full collection load
- Add StreamingResponse for JSON export path using a generator that yields one document at a time, consistent with existing CSV streaming
- Add ISO date format validation with clear 400 error messages
- Preserve existing API response structure and query parameters

### DIFF
--- a/backend/app/routes/advanced.py
+++ b/backend/app/routes/advanced.py
@@ -1,8 +1,9 @@
 from datetime import datetime
+import json
 from fastapi import APIRouter, Depends, Query, HTTPException
 from app.services.summarizer import generate_daily_summary, extract_key_insights
 from app.services.timeline import get_today_timeline
-from app.services.memory_store import get_memory_stats, all_memories
+from app.services.memory_store import get_memory_stats, filtered_memories
 from app.services.auth import get_current_user_id
 from app.services.memory_graph import build_memory_graph
 from app.core.logging_config import logger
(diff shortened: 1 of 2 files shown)


KaparthyReddy:
The ruff CI failure is a pre-existing configuration issue in `ruff.toml` (invalid `line-length` field for this version of ruff) — unrelated to this PR. These changes are limited to `backend/app/routes/advanced.py` and `backend/app/services/memory_store.py`.

Chetan0e:
@KaparthyReddy 

This PR looks good overall and properly addresses the main export performance bottleneck by moving filtering to the MongoDB query layer instead of filtering large datasets in Python memory.
The JSON streaming approach, date validation handling, and scoped implementation are all clean and well structured.
One small note:
`filtered_memories()` still builds an intermediate Python list before the response generator streams the output, so this is not fully end-to-end streaming yet for extremely large exports. Not a blocker for this PR since the primary query/memory inefficiency has already been significantly improved.
Nice work keeping the changes focused and production-oriented.
i am merging the pr , great work

PR #42, opened 2026-05-22
fix: move WebSocket JWT auth from URL query string to first-message frame

### LINKED ISSUE #35: Security: JWT exposed in WebSocket query string
**Describe the bug**
The `/ws/updates` WebSocket endpoint accepts the JWT as a URL query parameter (`?token=...`). Tokens passed this way are logged in full by most web servers, reverse proxies, and CDNs by default, and may also persist in browser history and leak via `Referer` headers — exposing valid session tokens to anyone with access to those logs.

**To Reproduce**
1. Establish a WebSocket connection to `/ws/updates?token=<valid_jwt>`
2. Check your server or reverse proxy access logs
3. Observe the full JWT is recorded in the logged URL

**Expected behavior**
The token should never appear in the URL. It should be passed either via the `Sec-WebSocket-Protocol` header during the HTTP upgrade handshake, or as a JSON auth frame sent immediately after the connection is accepted — both keep the token out of logs and browser history.

**Screenshots**
N/A

**Desktop (please complete the following information):**
- OS: Any
- Browser: Any
- Version: Any

**Smartphone (please complete the following information):**
- Device: Any
- OS: Any
- Browser: Any
- Version: Any

**Additional context**
Affected file: `backend/app/routes/websocket.py` → `websocket_endpoint()`

```python
@router.websocket("/ws/updates")
async def websocket_endpoint(websocket: WebSocket, token: str):
```

The `token` parameter is bound directly from the query string by FastAPI. Two common fixes:
- **Option A — `Sec-WebSocket-Protocol` header:** client sends the JWT as a subprotocol value during the HTTP upgrade; server reads it from `websocket.headers` before calling `websocket.accept()`
- **Option B — first-message auth:** accept the connection, then require a `{"type": "auth", "token": "..."}` frame as the very first message; close with code `4001` if it is missing or invalid within a short timeout

Vishakhayadavv:
Closes #35

## Problem
`/ws/updates` accepted JWT via `?token=...` query parameter.
Tokens in URLs are logged by web servers, reverse proxies, and CDNs,
and may persist in browser history.

## Fix
- Removed `token: str` query parameter from `websocket_endpoint()`
- Added `import asyncio`
- Server now accepts the connection first, then waits up to 10 seconds
  for a `{"type": "auth", "token": "<jwt>"}` JSON frame
- Closes with code `4001` if frame is missing, malformed, or token is invalid

## No frontend changes needed
No WebSocket client code exists in `web/src/` that references this endpoint.


Chetan0e (changes requested):
@Vishakhayadavv 
Thanks for the contribution and security improvement effort 
The overall direction of moving WebSocket JWT auth away from URL query parameters is valid and improves token exposure safety.

Before merging, a few adjustments are needed:

* Please retain backward compatibility temporarily by supporting both query-string auth and first-message auth during migration.
* Keep connection registration routed through `manager.connect()` instead of directly mutating `manager.active_connections`.
* Ensure the updated auth flow is documented clearly for future frontend/client integrations.

The speaker route authentication additions look good 
Please update the PR accordingly and I’ll review again.

Chetan0e:
@Vishakhayadavv Thanks for the contribution and security improvement effort
everything look perfect keep it up , i am merging this pr

PR #44, opened 2026-05-22
fix: correct @cached decorator order on /insights route

### LINKED ISSUE #36: cached decorator order is wrong on /insights — caching bypassed
**Describe the bug**
On the `/insights` route, the `@cached()` decorator is applied in the wrong order relative to `@router.get()`. Every other cached route in the file puts `@router.get()` first, then `@cached()` — but `/insights` has them reversed. FastAPI registers the raw function before the cache wrapper is applied, so the decorator silently has no effect and every request hits the LLM regardless of TTL.

**To Reproduce**
1. Make a `GET /insights` request
2. Make the same request again within the 15-minute TTL window
3. Observe the response time is identical — no cache hit occurs and the underlying service is called every time

**Expected behavior**
Repeated calls to `GET /insights` within the 15-minute TTL window should be served from cache, consistent with the behaviour of `/summary`, `/statistics`, and `/graph`.

**Screenshots**
N/A

**Desktop (please complete the following information):**
- OS: Any
- Browser: Any
- Version: Any

**Smartphone (please complete the following information):**
- Device: Any
- OS: Any
- Browser: Any
- Version: Any

**Additional context**
Affected file: `backend/app/routes/advanced.py` → `insights()` (~line 55)

Current (broken) order:
```python
@cached(ttl_seconds=900, key_prefix="insights")
@router.get("/insights")
async def insights(user_id: str = Depends(get_current_user_id)):
```

Expected order (matching all other cached routes in the file):
```python
@router.get("/insights")
@cached(ttl_seconds=900, key_prefix="insights")
async def insights(user_id: str = Depends(get_current_user_id)):
```

Vishakhayadavv:
Closes #36

## Problem
`@cached()` was applied before `@router.get()` on the `/insights` route,
so FastAPI registered the raw function before the cache wrapper — making
the cache silently have no effect on every request.

## Fix
Swapped the two decorator lines to match the correct order used by all
other cached routes in the file (`/summary`, `/statistics`, `/graph`):


Chetan0e:
@Vishakhayadavv great job

Vishakhayadavv:
thank you!

Issue #51, opened 2026-05-23
[Bug] Logout is cosmetic — blacklisted tokens remain valid for up to 7 days


Vishakhayadavv:
**Describe the bug**
`/auth/logout` correctly writes the token's `jti` to the `blacklisted_tokens` collection, but `verify_access_token()` never queries that collection. A logged-out token continues to authenticate every protected endpoint until it naturally expires — up to 7 days later.

**To Reproduce**
1. Log in and receive an `access_token`
2. Call `POST /auth/logout` with that token
3. Immediately call any authenticated endpoint (e.g. `GET /query`) using the same token
4. Observe a `200 OK` — the token still works despite being logged out

**Expected behavior**
Any token whose `jti` appears in `blacklisted_tokens` should be rejected with `401 Unauthorized` immediately after logout, regardless of its expiry time.

**Screenshots**
N/A

**Desktop (please complete the following information):**
- OS: Any
- Browser: Any
- Version: Any

**Smartphone (please complete the following information):**
- Device: Any
- OS: Any
- Browser: Any
- Version: Any

**Additional context**
Affected file: `backend/app/services/auth.py` → `verify_access_token()`

Current (broken) — blacklist is never consulted:
```python
def verify_access_token(token: str) -> Optional[str]:
    try:
        payload = jwt.decode(token, settings.secret_key, algorithms=[ALGORITHM])
        if payload.get("type") != "access":
            return None
        return payload.get("sub")
    except JWTError:
        return None
```

Fix — check the blacklist before returning the username:
```python
async def verify_access_token(token: str) -> Optional[str]:
    try:
        payload = jwt.decode(token, settings.secret_key, algorithms=[ALGORITHM])
        if payload.get("type") != "access":
            return None
        jti = payload.get("jti")
        if jti:
            db = get_db()
            if await db["blacklisted_tokens"].find_one({"jti": jti}):
                return None
        return payload.get("sub")
    except JWTError:
        return None
```

Note: this also requires updating all callers of `verify_access_token` to `await` the result, since it becomes a coroutine. The `get_current_user_id` dependency in `auth.py` and the equivalent in `reminders.py` both call it synchronously and will need updating too.

Vishakhayadavv:
kindly assign this issue to me under gssoc'26

Chetan0e:
Thanks for raising and documenting the issue clearly @Vishakhayadavv  — the report and root cause analysis were very helpful.

[...]

PR #56, opened 2026-05-25
fixed black listed token validity

Aayannoori15:
this PR points to issue #51 . Logout now actually revokes the current access token instead of just recording it. We updated token verification to check the blacklist on every authenticated request, so once a user logs out, that token is denied immediately rather than continuing to work until its 7-day expiry. The async auth call sites were updated as part of that change, and a regression test was added to make sure blacklisted tokens return 401 Unauthorized.

### DIFF
--- a/backend/app/routes/query.py
+++ b/backend/app/routes/query.py
@@ -2,7 +2,7 @@ import logging
 from fastapi import APIRouter, Depends, HTTPException, Query
 from app.services.query_engine import run_query
 from app.models.schema import QueryResponse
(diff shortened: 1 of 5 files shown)


Chetan0e:
Thanks for the contribution @Aayannoori15 — this PR correctly fixes the token blacklist validation issue and includes proper async updates along with regression test coverage.

Since the implementation is solid and resolves the security issue as expected, I’m proceeding with merging this PR.

Appreciate the contribution — feel free to work on more open issues in the project.

PR #29, opened 2026-05-21
fix: improve login card visibility and accessibility

jyoti205:
### LoginScreen.js
- Increased input background opacity (0.6 → 0.85)
- Improved input border visibility (0.05 → 0.15 opacity)
- Enhanced icon and text contrast (#64748b → #94a3b8)
- Added focus state with blue border highlight (#38bdf8)
- Added form card wrapper for better visual grouping

### RegisterScreen.js
- Same visibility and contrast improvements
- Added focus state with purple border highlight (#818cf8)
- Improved eye icon contrast

Fixes #18

PR #59, opened 2026-05-26
fix: register AppState listener in setupOfflineQueueListener

### LINKED ISSUE #48: [Bug] setupOfflineQueueListener never registers AppState listener — auto drain on foreground silently broken
## Bug Description
`setupOfflineQueueListener()` is exported and intended to automatically 
drain the offline queue when the app returns to foreground. However, the 
`AppState.addEventListener` call is commented out, making this feature 
completely non-functional with no warning to the caller.

## Steps to Reproduce
1. Call `setupOfflineQueueListener(axiosInstance)` in your app entry point
2. Add some requests to the offline queue
3. Background the app
4. Restore network connection
5. Bring app back to foreground
6. Observe — queue is never drained automatically

## Expected Behavior
Queue should automatically drain when app comes to foreground, since 
`setupOfflineQueueListener` is a public API that implies this behavior.

## Actual Behavior
The AppState listener is never registered — the function just returns 
a handler without attaching it anywhere:

```js
// AppState.addEventListener('change', handleAppStateChange);
// ↑ commented out — listener never fires
```

## Impact
- Developers who call `setupOfflineQueueListener()` expect auto-drain behavior
- Queue silently never drains unless `drainQueue()` is manually called
- No error or warning is thrown to indicate the feature is inactive

## Suggested Fix
Either:
1. Uncomment and properly implement the AppState listener, or
2. Remove the export and add clear docs that `drainQueue()` must be called manually

## Environment
- File: `mobile/services/offlineQueue.js`
- Function: `setupOfflineQueueListener()`

jyoti205:
## Summary

Fixes the inactive offline queue foreground listener in `setupOfflineQueueListener()`.

Previously, the `AppState` listener was not properly registered, causing automatic queue draining to never trigger when the app returned to the foreground.

## Changes Made

* Registered `AppState.addEventListener('change', handleAppStateChange)`
* Added proper cleanup using `subscription.remove()`
* Ensured queue drains automatically when app becomes active

## Issue Fixed

Closes #48

## Testing

Tested by:

1. Adding requests to offline queue
2. Backgrounding the app
3. Restoring network connection
4. Bringing app back to foreground

Queue now drains automatically as expected.

### DIFF
--- a/mobile/screens/LoginScreen.js
+++ b/mobile/screens/LoginScreen.js
@@ -1,15 +1,8 @@
 import React, { useState } from "react";
 import { 
-  View, 
-  Text, 
-  StyleSheet, 
-  TextInput, 
-  TouchableOpacity, 
-  KeyboardAvoidingView, 
-  Platform, 
-  Dimensions, 
-  ActivityIndicator,
-  Alert
+  View, Text, StyleSheet, TextInput, TouchableOpacity, 
+  KeyboardAvoidingView, Platform, Dimensions, 
+  ActivityIndicator, Alert
 } from "react-native";
 import { LinearGradient } from "expo-linear-gradient";
 import { MaterialCommunityIcons } from "@expo/vector-icons";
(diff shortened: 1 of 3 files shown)


Chetan0e:
@jyoti205 Thanks for the fix — the AppState listener registration and cleanup look correct and should resolve the offline queue foreground sync issue.

However, this PR also includes extensive UI/UX updates to LoginScreen and RegisterScreen that are unrelated to issue #48. To keep the PR focused and easier to review/test, please move the auth screen styling/accessibility changes into a separate PR.

Also, please verify that the cleanup function returned from `setupOfflineQueueListener()` is actually used by the caller (e.g. inside a `useEffect` cleanup) to avoid accumulating listeners during re-renders/remounts.

Once the unrelated UI changes are separated, the queue listener fix itself looks good.

jyoti205:
Hi @Chetan0e! I've separated the UI changes from this PR the login/accessibility fixes were already merged in PR #29. This PR now only contains the AppState listener fix 

the current CI failure is also unrelated to my changes.. it's a backend ImportError where get_current_user cannot be imported from app.services.auth which appears to be a pre-existing issue in the repo.

Chetan0e:
@jyoti205 Thanks for addressing the review feedback and separating the unrelated UI changes from this PR 
The PR is now clean, focused, and the AppState listener registration/cleanup flow looks correct for the offline queue synchronization behavior.
The remaining CI issue appears unrelated to these changes.
Merging this PR — thanks for the contribution

PR #62, opened 2026-05-26
fix: prevent pending requests from hanging on refresh failure

### LINKED ISSUE #61: Pending requests can hang indefinitely when token refresh fails
## Bug Description

When multiple requests encounter an expired access token simultaneously, pending requests subscribe to the token refresh flow through `refreshSubscribers`.

If the refresh request fails (expired/invalid refresh token, network error, etc.), queued subscriber callbacks are never rejected or cleared.

As a result, pending requests can remain unresolved indefinitely.

## Affected File

* `mobile/services/api.js`

## Root Cause

The refresh flow resolves subscribers on successful refresh using `onRefreshed(token)`, but there is no failure path that rejects or clears pending subscribers when refresh fails.

## Expected Behavior

Queued requests should:

* fail gracefully when refresh fails
* reject pending promises properly
* clear stale subscriber callbacks

## Proposed Fix

Add a failure handler for refresh subscribers that:

* rejects waiting requests on refresh failure
* clears subscriber queue on both success and failure
* preserves existing refresh architecture

I’d like to work on this issue.

hitdepani:
# Fix mobile token refresh queue edge case

Closes #61

## Summary
Fixes an edge case in the mobile token refresh queue flow where pending requests could remain unresolved if the token refresh failed.

## Changes
- Added failure handling for refresh subscribers.
- Configured the system to reject queued requests when a refresh fails.
- Cleared the subscriber queue on both success and failure scenarios.
- Preserved the existing refresh architecture.
- Ensured the `isRefreshing` state resets correctly.

## Result
Pending requests now fail gracefully instead of hanging indefinitely after a refresh failure.

### DIFF
--- a/mobile/services/api.js
+++ b/mobile/services/api.js
@@ -17,11 +17,16 @@ function subscribeTokenRefresh(callback) {
   refreshSubscribers.push(callback);
 }
 
+function rejectTokenRefresh(error) {
+  refreshSubscribers.forEach((callback) => callback(null, error));
+  refreshSubscribers = [];
+}
+
 /**
  * Execute all pending callbacks with new token
  */
 function onRefreshed(token) {
-  refreshSubscribers.forEach(callback => callback(token));
+  refreshSubscribers.forEach(callback => callback(token, null));
   refreshSubscribers = [];
 }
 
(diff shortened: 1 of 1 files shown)


hitdepani:
@Chetan0e  I've implemented the requested fix and ensured pending subscribers are properly rejected and cleared on refresh failure while preserving the existing flow. Please review when convenient. Thanks!

Chetan0e:
@hitdepani Thanks for the contribution 

The fix looks good and properly addresses the refresh queue edge case. Rejecting queued requests on refresh failure and clearing subscribers in both success/failure paths is the correct approach here.

I also appreciate that the existing refresh architecture was preserved while fixing the hanging-request behavior.

Merging this PR — great work

hitdepani:
@Chetan0e Thanks!!

PR #64, opened 2026-05-26
fix: restore backward compatibility for legacy pipeline and query APIs

### LINKED ISSUE #63: Regression: auth refactor removed get_current_user dependency, breaking route imports and backend startup
## Description

The recent auth dependency refactor introduced a backward-compatibility regression by removing or renaming the previously exported `get_current_user` dependency from `app.services.auth`.

Several routes (for example `backend/app/routes/speaker.py`) still import:

```python
from ..services.auth import get_current_user
```

However, `auth.py` now only exposes:

```python
get_current_user_id
```

As a result, application startup and pytest collection fail with:

```text
ImportError: cannot import name 'get_current_user' from 'app.services.auth'
```

---

## Impact

This issue currently causes:

- backend startup failures
- pytest collection failures
- broken route imports
- CI instability
- compatibility regressions for older route handlers depending on the previous auth dependency contract

---

## Suggested Fix

Restore backward compatibility by either:

- reintroducing a compatibility wrapper for `get_current_user`, OR
- updating all dependent routes/imports consistently across the codebase

A compatibility wrapper may be preferable to minimize blast radius and avoid cascading regressions in older modules/tests.

---

## Affected Areas

- `backend/app/services/auth.py`
- `backend/app/routes/speaker.py`
- potentially additional routes/tests importing `get_current_user`

---

## Reproduction

Run:

```bash
pytest
```

Current failure:

```text
ImportError: cannot import name 'get_current_user' from 'app.services.auth'
```

GOKULRAM-K:
Closes #63 

# Summary

This PR restores several legacy compatibility contracts that were unintentionally broken during recent service and pipeline refactors.

The repository currently contains older tests/routes/integrations that still depend on previously exposed functional APIs and module paths. Recent architectural changes migrated many implementations internally without preserving backward-compatible exports.

This PR adds lightweight compatibility layers and wrapper exports to stabilize the existing integration surface while preserving the newer architecture.

---

# Changes

## Compatibility Module Shims

Added compatibility wrapper modules for legacy imports:

- `app.services.embedding`
- `app.services.llm`

These now delegate to the current implementation modules.

---

## Restored Functional Pipeline APIs

Reintroduced backward-compatible helper exports for:

- `validate_text`
- `determine_lifecycle`
- `detect_correction`
- `parse_temporal_expression`
- `classify_intent`
- `extract_entities`

These preserve the older functional API contracts expected by existing tests and integrations.

---

## Query API Compatibility

Restored legacy `confidence` response field alongside the newer `confidence_score` field in `/query` responses to maintain backward compatibility.

---

## Timeline Compatibility

Added compatibility wrapper for legacy timeline access patterns.

---

## Additional Improvements

- improved graceful degradation when embedding generation is unavailable
- reduced query pipeline hard-failure behavior in offline/CI environments
- stabilized query-related tests and compatibility flows

---

# Why This Matters

These compatibility regressions currently cause:

- pytest collection failures
- broken imports
- stale integration failures
- API contract drift
- CI instability

This PR reduces regression surface while allowing the refactored architecture to remain intact.

---

# Validation

Validated locally with:

```bash
pytest tests/test_query.py -vv
```

Result:

```text
4 passed
```

### DIFF
--- a/backend/app/db/memory_lifecycle.py
+++ b/backend/app/db/memory_lifecycle.py
@@ -187,3 +187,17 @@ class MemoryLifecycleManager:
             return {"short_term": 0, "long_term": 0, "archived": 0}
 
 memory_lifecycle_manager = MemoryLifecycleManager()
+
+# ── Backward compatibility helper export ────────────────────────────────────
+
+def determine_lifecycle(importance: float) -> str:
+    """
+    Backward-compatible lifecycle determination helper.
+
+    Preserves older lifecycle classification API expected by tests
+    and integrations.
+    """
+    if importance >= MemoryLifecycleManager.IMPORTANCE_THRESHOLD:
+        return "long_term"
+
+    return "short_term"
--- a/backend/app/models/schema.py
+++ b/backend/app/models/schema.py
@@ -55,6 +55,7 @@ class QueryResponse(BaseModel):
             "0.0 = very uncertain, 1.0 = highly confident."
         ),
     )
+    confidence: float | None = None
 
 
 # ── Pipeline ──────────────────────────────────────────────────────────────────
--- a/backend/app/pipeline/data_validator.py
+++ b/backend/app/pipeline/data_validator.py
@@ -139,3 +139,20 @@ class DataValidator:
         return text.strip()
 
 data_validator = DataValidator()
+
+
+def validate_text(text: str):
+    """
+    Backward-compatible validation helper expected by legacy tests.
+    """
+
+    sanitized = data_validator.sanitize_input(text)
+
+    is_valid = (
+        data_validator._is_valid_length(sanitized)
+        and not data_validator._is_noise(sanitized)
+    )
+
+    reason = "valid" if is_valid else "invalid"
+
+    return is_valid, reason
--- a/backend/app/pipeline/extraction_pipeline.py
+++ b/backend/app/pipeline/extraction_pipeline.py
@@ -321,3 +321,71 @@ Format as JSON:
         return min(boost, 0.5)  # Cap at 0.5
 
 extraction_pipeline = ExtractionPipeline()
+
+# ── Backward compatibility helper exports ───────────────────────────────────
+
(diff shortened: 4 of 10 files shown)


Chetan0e:
@GOKULRAM-K Thanks for the contribution — the compatibility restoration work is valuable and the overall direction makes sense, especially the wrapper exports and legacy API preservation.

However, I noticed a few issues that should be cleaned up before merge:

Several helper functions appear duplicated in extraction_pipeline.py
determine_lifecycle() is duplicated in memory_lifecycle.py
auth.py may now be incompatible with the recent async verify_access_token() changes from the blacklist-token fix PR and likely needs await verify_access_token(...)

The compatibility shims themselves look useful, and the /query backward compatibility restoration is reasonable, but the duplicate exports and potential async regression should be resolved first.

Please clean those up and rebase against the latest main, then I’ll review again

GOKULRAM-K:
Thanks for the review.

I cleaned up the duplicated compatibility helper exports, removed the duplicate `determine_lifecycle()` definition, rebased against the latest `main`, aligned the auth compatibility layer with the recent async token verification changes, and restored the remaining missing compatibility wrappers/signatures expected by older extraction pipeline integrations.

I also revalidated the compatibility/query flow locally after rebasing (`tests/test_query.py` passes locally).

The remaining failing tests now appear to be unrelated behavioral/business-logic drift rather than compatibility-contract regressions.

Chetan0e:
@GOKULRAM-K Thanks for addressing the review feedback.

I've verified that the duplicate compatibility exports have been cleaned up, the lifecycle helper duplication has been removed, and the auth compatibility layer has been updated to align with the recent async token verification changes.

The compatibility wrappers and legacy API restorations now look consistent with the goal of preserving older integrations while keeping the newer architecture intact.

The remaining test failures appear unrelated to the compatibility-contract regressions addressed in this PR.

Approved. Thanks for the contribution

PR #65, opened 2026-05-26
fix: strict validation and standardized 422 for reminders hours param

### LINKED ISSUE #24: bug: @cached decorator ordering causes cross-user data leakage on /insights and /statistics; cache admin endpoints lack authorization
### Bug 1 — Wrong decorator order causes shared cache across users (Critical)

In `backend/app/routes/advanced.py`, two routes have the `@cached` decorator applied in the wrong order relative to `@router.get`:

**`/insights`** — `@cached` is above `@router.get`:
```python
@cached(ttl_seconds=900, key_prefix="insights")
@router.get("/insights")
async def insights(user_id: str = Depends(get_current_user_id)):
```

**`/statistics`** — `@cached` is below `@router.get`:
```python
@router.get("/statistics")
@cached(ttl_seconds=300, key_prefix="stats")
async def statistics(user_id: str = Depends(get_current_user_id)):
```

FastAPI requires `@router.get` to be the outermost decorator. When `@cached` wraps the function before FastAPI registers it, the `Depends(get_current_user_id)` injection does not execute during cache key generation. This means the cache key is built **without the user_id**, so all users share the same cached response.

**Impact:** User A's private memory insights/statistics are returned to User B if they hit the endpoint within the cache TTL window. This is a **privacy data leak**.

### Bug 2 — Cache admin endpoints have no authorization (Medium)

`GET /cache/stats` and `POST /cache/invalidate` are labeled as admin endpoints in comments but have no role check:

```python
@router.post("/cache/invalidate")
async def invalidate_user_cache(user_id: str = Depends(get_current_user_id)):
    invalidate_cache(pattern=user_id)  # any user can call this

@router.get("/cache/stats")
async def cache_stats():  # no auth at all
    return get_cache_stats()
```

Any authenticated user can view platform-wide cache statistics or invalidate any other user's cache by manipulating the request.

### Proposed Fix

1. Fix decorator order on `/insights` and `/statistics` so `@router.get` is always outermost and the cache key includes `user_id`
2. Add admin role verification middleware or a role check to `/cache/stats` and `/cache/invalidate`

### Area
`backend/app/routes/advanced.py`

uppalriya371-blip:
## What does this PR do?
Adds explicit server-side validation to the `hours` query parameter 
in `GET /reminders/upcoming`.

## Problem
The `hours` parameter had `ge=1, le=168` but no standardized error 
response. Invalid values like 0, -1, or 9999 didn't return clear, 
structured error messages.

## Fix
- Added explicit `if not (1 <= hours <= 168)` check
- Returns structured 422 with `field`, `message`, and `received` value
- Updated Query description to mention valid range
- Added tests for invalid (0, -1, 169, 9999, abc) and valid (1, 24, 168) inputs

Fixes #24

### DIFF
--- a/backend/app/routes/reminders.py
+++ b/backend/app/routes/reminders.py
@@ -58,6 +58,16 @@ async def get_upcoming(
     Return all pending reminder alerts for the current user
     within the next `hours` hours (default 24).
     """
+    if not (1 <= hours <= 168):
+        raise HTTPException(
+            status_code=422,
+            detail={
+                "error": "Validation failed",
+                "message": "hours must be between 1 and 168",
+                "field": "hours",
+                "received": hours
+            }
+        )
     reminders = await get_upcoming_reminders(
         user_id=user,
         hours=hours,


Chetan0e:
@uppalriya371-blip Thanks for the contribution 

FastAPI already performs range validation for the `hours` parameter through the existing `Query(..., ge=1, le=168)` constraints, so the main value of this PR is the improved standardized error structure and clearer validation messaging.

The structured 422 response format and additional test coverage are useful improvements for API consistency, so I’m proceeding with the merge.

Appreciate the contribution

PR #67, opened 2026-05-28
Persist privacy mode to MongoDB — survives server restarts 

### LINKED ISSUE #57: Privacy mode resets silently on server restart — _PRIVATE_MODES uses in-memory state
## Bug Description

`backend/app/services/privacy.py` stores privacy mode in a plain Python dict:

```python
_PRIVATE_MODES: dict[str, bool] = {}
```

This means **every server restart silently resets all users' privacy modes to `False`**, regardless of what they had previously set.

A user who explicitly enabled privacy mode (`POST /privacy/toggle`) to stop audio processing has no way of knowing their reference was lost. Their data starts being processed again without any notification or indication.

## To Reproduce

1. Enable privacy mode: `POST /privacy/toggle` → `{"private": true}`
2. Restart the server
3. Check privacy state: `GET /privacy` → `{"private": false}` ← silently reset

## Expected Behavior

Privacy mode should persist across server restarts. A user's explicit privacy preference should survive the process lifecycle.

## Proposed Fix

Persist privacy state to MongoDB — same database already used for user data:

- On `toggle_privacy(user_id)`: write state to `users` collection or a dedicated `privacy_settings` collection
- On `is_private(user_id)`: read from MongoDB with an in-memory cache layer for performance
- Keep the existing function signatures intact so no callers need to change

This is a focused, single-file change to `backend/app/services/privacy.py` with a corresponding update to `backend/tests/test_privacy.py`.

## Affected File

`backend/app/services/privacy.py` — `toggle_privacy()` and `is_private()`

## Additional Context

Privacy toggle authentication was fixed in #30/#31. This is the natural follow-up — the toggle is now protected, but the state it toggles doesn't survive restarts.

AjayBandiwaddar:
## Summary
Fixes the silent privacy mode reset on server restart by persisting state to MongoDB instead of relying on in-memory dict.

Closes #57

## Root Cause
`_PRIVATE_MODES` was a plain Python dict — wiped on every process restart. A user who enabled privacy mode had no indication their preference was lost.

## Changes
- `backend/app/services/privacy.py` — both functions converted to async;   state persisted to `users.privacy_enabled` field in MongoDB; in-memory cache retained as fast path, populated from MongoDB on cache miss
- `backend/app/routes/privacy.py` — added `await` to both calls
- `backend/app/services/pipeline.py` — added `await` to `is_private()` call
- `backend/tests/test_privacy.py` — existing tests updated for async; added `test_privacy_persists_after_cache_clear()` which clears `_PRIVATE_MODES` dict and verifies state is recovered from MongoDB - `backend/app/routes/speaker.py` — fixed broken import (`get_current_user` → `get_current_user_id`) that prevented server startup

## Verified Locally
1. Enable privacy → restart server → privacy still `True` ✅
2. All three existing privacy tests pass ✅
3. New persistence test passes ✅

### DIFF
--- a/backend/app/routes/privacy.py
+++ b/backend/app/routes/privacy.py
@@ -1,5 +1,4 @@
 from fastapi import APIRouter, Depends
-
 from app.services.auth import get_current_user_id
 from app.services.privacy import is_private, toggle_privacy
 
@@ -8,9 +7,9 @@ router = APIRouter(prefix="/privacy", tags=["privacy"])
 
 @router.get("/")
 async def get_privacy(user_id: str = Depends(get_current_user_id)):
-    return {"private": is_private(user_id)}
+    return {"private": await is_private(user_id)}
 
 
 @router.post("/toggle")
 async def toggle(user_id: str = Depends(get_current_user_id)):
-    return {"private": toggle_privacy(user_id)}
+    return {"private": await toggle_privacy(user_id)}
--- a/backend/app/routes/speaker.py
+++ b/backend/app/routes/speaker.py
@@ -2,17 +2,17 @@ from fastapi import APIRouter, Depends
 from app.models.schema import VoiceTrainRequest
 from app.services.gemini_embedding import get_embedding
 from app.services.speaker_training import add_voice, get_voice_profiles
-from ..services.auth import get_current_user
+from ..services.auth import get_current_user_id
 
 router = APIRouter()
 
 @router.post("/train")
-def train_voice(payload: VoiceTrainRequest, _: dict = Depends(get_current_user)):
+def train_voice(payload: VoiceTrainRequest, _: dict = Depends(get_current_user_id)):
     sample = payload.sample_text or payload.name
     embedding = get_embedding(sample)
     add_voice(payload.name, embedding)
     return {"msg": "voice profile saved", "name": payload.name}
 
 @router.get("/profiles")
-def list_profiles(_: dict = Depends(get_current_user)):
+def list_profiles(_: dict = Depends(get_current_user_id)):
     return {"profiles": get_voice_profiles()}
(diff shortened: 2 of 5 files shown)


AjayBandiwaddar:
@Chetan0e — worth flagging: `pipeline.py` line 22 calls `is_private(user_id)` synchronously. After converting `is_private` to async in this PR, that call is also updated to `await is_private(user_id)` — without this, the coroutine object would always evaluate as truthy and silently skip all audio processing regardless of the user's actual privacy setting.

AjayBandiwaddar:
@Chetan0e — the CI failures appear to be pre-existing regressions unrelated to this PR. Failures include missing exports from `app.pipeline.extraction_pipeline`, missing `app.services.embedding` and `app.services.llm` modules, and auth test failures from the async `verify_access_token` refactor — none touched by this PR.

This PR only modifies: `services/privacy.py`, `routes/privacy.py`, `services/pipeline.py` (one await added), `tests/test_privacy.py`, and `routes/speaker.py` (import fix).

Happy to investigate the broader CI failures separately if helpful.

Chetan0e:
@AjayBandiwaddar Thanks for the implementation. The MongoDB-backed persistence addresses the root cause described in #57, and the async updates to the privacy flow and pipeline call site look correct.

Before merging, I'd like clarification on a couple of points:

1. What happens if the target user document does not exist in MongoDB? I don't see an `upsert=True` path, and it's not clear whether user records are guaranteed to exist before privacy settings are toggled.

2. Could you add a test covering the missing-user scenario (or clarify why it cannot occur)? This would help verify that privacy persistence behaves correctly even when no existing document is found.

Once those points are clarified or covered by tests, the overall persistence approach looks good.

Chetan0e:
@AjayBandiwaddar are you ready

AjayBandiwaddar:
@Chetan0e — pushed the fix. Changed `upsert=False` to `upsert=True` in `toggle_privacy()` and added `test_toggle_privacy_for_nonexistent_user_persists_correctly` which toggles privacy for a user with no existing document, clears the in-memory cache, and verifies state is correctly recovered from MongoDB.

AjayBandiwaddar:
@Chetan0e Ready to merge whenever you are. All changes are tested and verified locally.

Chetan0e:
@AjayBandiwaddar cleared , merging this pr , thank you for contributing
keep it up

PR #68, opened 2026-05-28
fix: resolve splash screen indefinite hang by decoupling SplashScreen.hideAsync() from onLayout (#47)

### LINKED ISSUE #47: [BUG/OPTIMISATION]: Splash Screen Can Hang Indefinitely Due to Missing onLayout Trigger During Loading State
The app calls:

SplashScreen.preventAutoHideAsync();

But the splash screen is only hidden inside:

onLayout={onLayoutRootView}

attached to:

<View style={styles.container} onLayout={onLayoutRootView}>

The problem:

During the loading state, the component returns:

<View style={styles.loadingContainer}>

WITHOUT onLayout={onLayoutRootView}.

That means if fonts/auth take time or fail in certain edge cases, SplashScreen.hideAsync() may never execute correctly, causing the splash screen to remain visible indefinitely on some devices/builds.

Why This Matters
It’s an app startup reliability issue.

Symptoms users may face:

App frozen on splash screen
Blank startup on slow devices
Inconsistent startup behavior between Android/iOS
Production-only failures due to async race conditions

Tech4Aditya:
## Problem
Fixes #47

SplashScreen.hideAsync() was only called inside onLayoutRootView, 
which was attached to the root View. During the loading state, the 
component returns a different View WITHOUT onLayout attached — so 
hideAsync() never fires if fonts/auth take too long or fail, causing 
the splash screen to hang indefinitely.

## Fix
Moved SplashScreen.hideAsync() into a useEffect that watches 
fontsLoaded, fontError, and authChecked directly.

- Removed onLayoutRootView useCallback entirely
- Removed onLayout prop from all root Views
- hideAsync() now fires as soon as state resolves — not on layout mount
- fontError is handled so font load failures don't also hang the splash

## Changes
- mobile/App.js only, no other files touched.

### DIFF
--- a/mobile/App.js
+++ b/mobile/App.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback } from "react";
+import React, { useState, useEffect } from "react";
 import { NavigationContainer } from "@react-navigation/native";
 import AsyncStorage from "@react-native-async-storage/async-storage";
 import * as SplashScreen from "expo-splash-screen";
@@ -16,7 +16,7 @@ import Tabs from "./screens/Tabs";
 import LoginScreen from "./screens/LoginScreen";
 import RegisterScreen from "./screens/RegisterScreen";
 import { View, ActivityIndicator, Text, StyleSheet } from "react-native";
-import { colors, fonts } from "./theme";
+import { colors } from "./theme";
 
 // Keep splash screen visible while loading
 SplashScreen.preventAutoHideAsync();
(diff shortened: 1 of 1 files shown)


Chetan0e:
@Tech4Aditya Thanks for the fix. The root cause analysis is accurate and the implementation correctly decouples splash screen dismissal from the root view layout lifecycle.

Moving SplashScreen.hideAsync() into a readiness-based useEffect prevents hangs when the loading screen is rendered before the root view with onLayout mounts, and the additional handling for fontError and auth completion covers the failure paths as well.

The change is minimal, focused, and addresses the reported issue directly.

Approved.

PR #77, opened 2026-05-31
fix(memory_store): store_memories_batch returns mem_ids on success

### LINKED ISSUE #70: bug: `store_memories_batch` returns `None` instead of stored memory IDs
`store_memories_batch()` in `backend/app/services/memory_store.py` contains unreachable code that silently causes the function to return `None` on every successful batch write.

### Code path

```python
# lines 182–186
except Exception as e:
    ...
    raise          # ← line 183

    logger.info(f"Batch stored {len(items)} memories for user {user_id}")
    return mem_ids  # ← unreachable: placed after `raise`
```

On the success path — where no exception is thrown — execution exits the `try/except` block after the ChromaDB `upsert` and falls off the end of the function, returning `None` implicitly. The `return mem_ids` on line 186 is dead code.

### Impact

Any caller that uses the returned IDs for downstream linking, deduplication, or audit will receive `None`. This is a silent data-integrity failure: memories are written to both MongoDB and ChromaDB, but callers lose all references to them.

### Affected file

`backend/app/services/memory_store.py`, lines 160–186

### Reproduction

Call `store_memories_batch()` with two or more items and observe the return value is `None` rather than a list of UUIDs.

Srejoye:
## Problem

`store_memories_batch()` in `backend/app/services/memory_store.py` was silently returning `None` on every successful batch write.

The `logger.info(...)` and `return mem_ids` lines were placed after a bare `raise` inside the `except` block — syntactically valid Python but logically dead code. On the success path, execution fell off the end of the `try/except` and the function implicitly returned `None`.

Any caller relying on the returned IDs for downstream linking, deduplication, or audit received `None` with no error or warning.

## Fix

Moved `logger.info(...)` and `return mem_ids` outside the `except` block so they execute on the success path only. One indentation change, no logic altered.

## Tests

Added `tests/test_store_memories_batch.py` with 3 tests:

- `test_returns_list_of_uuids_on_success` — asserts return value is a list of N UUID strings, not `None`
- `test_returned_ids_match_mongodb_insert` — asserts returned IDs match the `_id` values actually inserted into MongoDB
- `test_rollback_on_chromadb_failure` — asserts ChromaDB failure triggers `delete_many` rollback and propagates the exception

All 3 pass locally.

## Related

Closes #70

### DIFF
--- a/backend/app/services/memory_store.py
+++ b/backend/app/services/memory_store.py
@@ -174,16 +174,13 @@ async def store_memories_batch(
             f"Batch ChromaDB upsert failed for user {user_id}, "
             f"rolling back MongoDB batch insert: {e}"
         )
-
-        # Rollback inserted Mongo documents
         await _memories_collection().delete_many({
             "_id": {"$in": mem_ids}
         })
-
         raise
 
-        logger.info(f"Batch stored {len(items)} memories for user {user_id}")
-        return mem_ids
+    logger.info(f"do i Batch stored {len(items)} memories for user {user_id}")
+    return mem_ids
 
 
 # ── Read / Search ─────────────────────────────────────────────────────────────
--- a/backend/tests/test_store_memories_batch.py
+++ b/backend/tests/test_store_memories_batch.py
@@ -1,0 +1,122 @@
+import pytest
+from unittest.mock import AsyncMock, MagicMock, patch
+
+
+class TestStoreMemoriesBatch:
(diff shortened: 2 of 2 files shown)


Chetan0e:
@Srejoye Thanks for the fix and the accompanying regression tests. The implementation addresses the root cause reported in #70 and the added test coverage is valuable. The changes are well scoped and look good to me.

Approved and merging.

PR #78, opened 2026-05-31
Configure mobile API URL via Expo env

### LINKED ISSUE #66: Mobile app BASE_URL should be configurable via environment or config file, not hardcoded
The mobile app currently requires every developer to manually edit a source file to set the backend URL before they can run the project. This is a friction point for onboarding new contributors and a common source of accidental commits with local IP addresses. The fix is to adopt Expo's standard environment variable support, which is already built into the framework and requires minimal changes.

Current behaviour (problem)
The README instructs contributors to open mobile/services/api.js and directly edit the source:
const BASE_URL = "http://YOUR_COMPUTER_IP:8000";
This creates multiple real problems for contributors and maintainers:

Every new contributor must modify a tracked file just to get the app running — it's the very first thing that breaks on a fresh clone.
Developers frequently forget and commit their local IP, polluting the git history with personal network details.
There's no clean way to switch between local, staging, and production environments without re-editing the file each time.
Expo EAS builds and CI pipelines have no mechanism to inject the correct URL at build time with the current setup.
It's inconsistent with how the backend already handles config — via .env and .env.example — so contributors must follow two different patterns

Aarushiprajapati:
## Summary

- Added `mobile/.env.example` with `EXPO_PUBLIC_API_URL` examples.
- Updated `mobile/config.js` to read the backend URL from Expo public env support.
- Added explicit gitignore coverage for local mobile env files.
- Updated README mobile setup and troubleshooting docs.

## Why

The mobile app had a committed backend URL fallback and docs told contributors to edit source files for local backend addresses. This made onboarding harder and risked accidental commits of personal LAN IPs.

## Impact

Contributors can now configure the mobile backend URL through `mobile/.env` using Expo's standard `EXPO_PUBLIC_*` pattern.

Fixes #66.

## Validation

- `git diff --check`
- `node --check mobile/config.js`
- Confirmed the previous hardcoded IP and README source-edit instructions are removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@
 .env.local
 .env.*.local
 .env.production
+mobile/.env
+mobile/.env.local
+mobile/.env.*.local
+mobile/.env.production
 # --- Python ---
 __pycache__/
 *.py[cod]
--- a/README.md
+++ b/README.md
@@ -463,14 +463,29 @@ npm run preview
 
 ### Step 6 — Mobile (optional)
 
-Edit `mobile/services/api.js`:
-
-```javascript
-const BASE_URL = "http://192.168.1.XXX:8000";  // your LAN IP, not localhost on device
-```
+Create the mobile environment file:
 
 ```bash
 cd mobile
+cp .env.example .env
+```
+
+Set `EXPO_PUBLIC_API_URL` in `mobile/.env` to the backend URL reachable from the target device:
+
+```env
+# Android emulator
+EXPO_PUBLIC_API_URL=http://10.0.2.2:8000
+
+# iOS simulator or web
+# EXPO_PUBLIC_API_URL=http://localhost:8000
+
+# Physical device on the same Wi-Fi
+# EXPO_PUBLIC_API_URL=http://YOUR_COMPUTER_IP:8000
+```
+
+Restart Expo after changing `.env`.
+
+```bash
 npm install
 npx expo start
 ```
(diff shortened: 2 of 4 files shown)


Chetan0e:
@Aarushiprajapati Thanks for the contribution!

The implementation looks good and addresses the onboarding/configuration issue described in #66. Moving the mobile API URL to Expo environment variables, adding a .env.example, updating the documentation, and preventing local env files from being committed are all valuable improvements.

One quick confirmation before merge:
Have you verified that all mobile API consumers are already using API_BASE_URL from mobile/config.js so the new environment-based configuration is picked up everywhere?

Other than that, the changes look clean and well scoped to the issue.

Aarushiprajapati:
Thanks for reviewing!

Yes, I verified this by checking the mobile API consumers after the change. The mobile request code now routes through `API_BASE_URL` from `mobile/config.js` either directly or through local constants derived from it:

- `mobile/services/api.js`
- `mobile/services/auth.js`
- `mobile/screens/LoginScreen.js`
- `mobile/screens/RegisterScreen.js`
- `mobile/screens/TimelineScreen.js`

I also checked for remaining hardcoded mobile API base URLs such as `localhost` / `127.0.0.1` in the mobile source and did not find any remaining mobile API consumer bypassing `mobile/config.js`.

`offlineQueue.js` receives the configured axios instance, so it also uses the same configured base URL indirectly.

Chetan0e:
@Aarushiprajapati Great i am merging this PR 
thank you for the contribution , keep it up

PR #79, opened 2026-05-31
perf: push export filters to MongoDB and stream JSON response in GET /export

### LINKED ISSUE #39: perf: GET /export fetches up to 10,000 memories into Python memory before filtering — memory exhaustion risk
## Description

In `backend/app/routes/advanced.py`, the `/export` endpoint has two performance problems:

### Problem 1 — Full collection load before filtering

```python
memories = await all_memories(user_id, limit=10000)

if intent_filter:
    memories = [m for m in memories if m.get("metadata", {}).get("intent") == intent_filter]

if start_date:
    start_dt = datetime.fromisoformat(start_date)
    memories = [m for m in memories if ...]
```

All 10,000 documents are fetched from MongoDB into Python memory first, then filtered in a loop. For users with large memory stores, this means:
- Up to 10,000 full documents held in RAM simultaneously
- Network transfer of all documents regardless of how many match the filter
- O(n) Python iteration on every request

The fix is to push `intent_filter` and date range into the MongoDB query itself using `$gte`/`$lte` on `created_at` and a metadata field match, so only matching documents are transferred.

### Problem 2 — JSON export has no streaming

The CSV path correctly uses `StreamingResponse`, but the JSON path builds the full list in memory and returns it as a single response body:

```python
return {
    "memories": memories,  # up to 10,000 documents in one response
    "count": len(memories),
    "exported_at": datetime.utcnow().isoformat()
}
```

For large exports this can exhaust server memory and cause the response to time out before the client receives anything.

## Suggested Fix
- Pass `intent_filter` and date range as MongoDB query filters before fetching
- Use `StreamingResponse` with `json.dumps` chunking for the JSON export path, consistent with the existing CSV implementation

## Area
`backend/app/routes/advanced.py` — `GET /export`

KaparthyReddy:
Fixes #39

### Problem 1 — Filter pushdown
`GET /export` was fetching up to 10,000 documents into Python memory before applying `intent_filter` and date range filters in a Python loop. For users with large memory stores this meant:
- Up to 10,000 full documents held in RAM per request
- Network transfer of all documents regardless of how many matched the filter
- O(n) Python iteration on every export request

Fixed by building a MongoDB query dict with `metadata.intent` and `created_at` `$gte`/`$lte` filters before fetching, so only matching documents are transferred from the database. Added `all_memories_filtered(query)` to `memory_store.py` to support arbitrary MongoDB queries.

### Problem 2 — JSON streaming
The CSV path correctly used `StreamingResponse`, but the JSON path built the full list in memory and returned it as a single response body — risking memory exhaustion and response timeouts on large exports.

Fixed by replacing the dict return with a `StreamingResponse` using a generator that yields JSON chunks, consistent with the existing CSV implementation.

## Files changed
- `backend/app/routes/advanced.py`
- `backend/app/services/memory_store.py`

### DIFF
--- a/backend/app/routes/advanced.py
+++ b/backend/app/routes/advanced.py
@@ -312,7 +312,7 @@ async def export_memories(
                     yield json.dumps(m, default=str)
                     if i < count - 1:
                         yield ","
-                yield f'], "count": {count}, "exported_at": "{exported_at}"}}\'
+                yield ']' + f', "count": {count}, "exported_at": "{exported_at}"}}'
 
             return StreamingResponse(
                 json_stream(),
--- a/backend/app/routes/advanced.py
+++ b/backend/app/routes/advanced.py
@@ -2,7 +2,7 @@ from datetime import datetime
 from fastapi import APIRouter, Depends, Query, HTTPException
 from app.services.summarizer import generate_daily_summary, extract_key_insights
 from app.services.timeline import get_today_timeline
-from app.services.memory_store import get_memory_stats, all_memories
+from app.services.memory_store import get_memory_stats, all_memories, all_memories_filtered
 from app.services.auth import get_current_user_id
 from app.services.memory_graph import build_memory_graph
 from app.core.logging_config import logger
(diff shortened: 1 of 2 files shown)


KaparthyReddy:
The CI test failures shown are pre-existing and unrelated to this PR's changes. The failures include rate limiting issues (429 vs 201), status endpoint mismatches ('running' vs 'ok'), and ObjectId validation errors — all present in the main branch before this PR. Our changes only affect backend/app/routes/advanced.py and backend/app/services/memory_store.py (filter pushdown and JSON streaming). Happy to rebase if needed.

Chetan0e:
@KaparthyReddy Thanks for the contribution. The MongoDB filter pushdown is a useful optimization and the changes are well scoped to the export flow. I've reviewed the implementation and am happy to merge this improvement. Thanks for contributing!

PR #105, opened 2026-06-05
fix(task-queue): make dead-letter retry atomic — prevent duplicate execution and stuck records

### LINKED ISSUE #73: bug: Non-atomic dead-letter retry produces duplicate task execution and permanent stuck records on partial failure
`retry_dead_letter_task()` in `backend/app/workers/task_queue.py` performs a non-atomic two-step operation: it inserts a task into the main queue before removing the dead-letter record. A crash between these steps can leave the system in an inconsistent state.

### Code path

```python
async def retry_dead_letter_task(self, task_id: str) -> bool:
    ...
    new_task = Task(task_id=task_id, ...)

    await self.enqueue(new_task)
    await db[self.DEAD_LETTER_COLLECTION].delete_one({"task_id": task_id})
```

### Failure modes

1. Crash between insert and delete → task exists in both queues
2. `DuplicateKeyError` gets swallowed inside `enqueue()` and the dead-letter record is still deleted

### Impact

Tasks may execute multiple times or become permanently lost depending on failure timing.

### Affected file

`backend/app/workers/task_queue.py`, lines 461–496

hariom888:
## Problem

Closes #73.

`retry_dead_letter_task()` performed a two-step, non-atomic operation:

```python
await self.enqueue(new_task)                                    # step 1
await db[self.DEAD_LETTER_COLLECTION].delete_one({"task_id"})  # step 2
```

Two failure modes:

1. **Crash between steps** → task exists in both queues; gets executed twice.
2. **`DuplicateKeyError` swallowed inside `enqueue()`** → dead-letter entry deleted despite the task never being re-queued (task loss).

`_move_to_dead_letter()` had the same risk in reverse (delete-before-insert).

### `retry_dead_letter_task`

- Generates a **fresh `task_id`** (UUID) for every retry, making re-runs idempotent and eliminating the `DuplicateKeyError` root cause entirely.
- Records `original_task_id` in the payload for operator audit.
- Checks `enqueue()` return value; **only deletes the dead-letter entry on confirmed success**. Returns `False` (preserving the dead-letter entry) if enqueue fails.

### `_move_to_dead_letter`

- **Inserts into dead-letter first**, then deletes from main queue.
- Treats a `DuplicateKeyError` on insert as an idempotency signal (prior crash already wrote the entry) and still removes the main-queue record.

## Failure-mode analysis after this PR

| Failure point | Outcome |
|---|---|
| Crash after enqueue, before dead-letter delete | Dead-letter entry remains; operator can retry again (new task_id prevents double-execution) |
| `enqueue()` returns `False` | Dead-letter entry untouched; no task loss |
| Crash after dead-letter insert, before main-queue delete | Task reachable via heartbeat reset; second `_move_to_dead_letter` sees `DuplicateKeyError`, logs warning, still cleans main queue |

## Testing

```bash
pytest backend/tests/test_dead_letter_retry.py -v
```

Covers: enqueue-failure preservation, new-task-id assertion, audit payload, duplicate retry idempotency, operation ordering, queue consistency pre/post.

### DIFF
--- a/backend/app/workers/task_queue.py
+++ b/backend/app/workers/task_queue.py
@@ -337,42 +337,65 @@ class TaskQueue:
             return False
     
     async def _move_to_dead_letter(self, task_id: str, error_message: str, stack_trace: str) -> bool:
-        """Move failed task to dead letter queue."""
+        """
+        Move a failed task to the dead-letter queue.
+
+        Ordering guarantee:
+        1. Insert into dead-letter collection.
+        2. Only if insert succeeded, delete from the main queue.
+
+        A crash between steps 1 and 2 leaves the task in both collections.
+        On the next dequeue cycle the heartbeat-timeout reset will make the
+        main-queue entry visible again; a worker will attempt to claim it and
+        immediately re-exhaust retries, reaching _move_to_dead_letter again.
+        The dead-letter insert will fail with DuplicateKeyError (unique index
+        on task_id), which is caught and logged — the entry is already safe.
+        """
         await self._ensure_initialized()
-        
+
         db = get_db()
         if db is None:
             logger.error("Database not available for dead letter queue")
             return False
-        
+
         try:
-            # Get task document with worker check
             task_doc = await db[self.COLLECTION_NAME].find_one({
                 "task_id": task_id,
                 "worker_id": self.worker_id
             })
             if not task_doc:
-                logger.error(f"Task {task_id} not found for dead letter queue or not owned by worker")
+                logger.error(
+                    f"Task {task_id} not found for dead letter queue or not owned by worker"
+                )
                 return False
-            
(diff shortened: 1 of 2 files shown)


Chetan0e:
@hariom888 Thanks for the contribution. The proposed changes address the dead-letter retry reliability issues and significantly improve failure recovery behavior.

The retry flow now preserves dead-letter entries when re-enqueueing fails, introduces fresh task IDs to avoid retry collisions, and adds useful auditability through `original_task_id`. The dead-letter migration ordering is also safer, reducing the risk of task loss during partial failures.

The accompanying test suite provides good coverage for the identified edge cases, including retry consistency, enqueue failures, duplicate retry scenarios, and operation ordering.

Changes look good to me. Approved for merge.

Issue #81, opened 2026-05-31
[BUG] POST /record/upload accepts any file type — no MIME type or extension validation allows non-audio files to be processed


KaparthyReddy:
In `backend/app/routes/record.py`, the `/record/upload` endpoint accepts any `UploadFile` with no content type or extension validation:

file: UploadFile = File(...)

A user can upload any file (PDF, image, executable) and it will be written to disk and passed to `process_audio()`. This wastes processing resources and may expose unexpected behaviour in the audio processing pipeline.

The endpoint should validate that the uploaded file has an audio MIME type (e.g. `audio/wav`, `audio/mpeg`, `audio/webm`) and/or a valid audio file extension before writing to disk.

KaparthyReddy:
Hi @Chetan0e — I'd like to work on this. I can add MIME type and file extension validation before writing to disk, rejecting non-audio files with a clear 400 error. Please assign this to me /assign gssoc26

madsysharma:
/claim
Hi @Chetan0e , I would like to work on this as part of GSSoC 2026.

Ashitha0409:
Hi @Chetan0e,

I'm Ashitha, a GSSoC 2026 contributor. I would like to work on this issue.
Could you please assign it to me?
Thank you.

Chetan0e:
Thanks for reporting this issue and providing a clear explanation of the impact.

The concern is valid. Since `/record/upload` is intended for audio processing, accepting arbitrary file types without validation can lead to unnecessary processing, wasted resources, and unexpected behavior in the audio pipeline.

Assigning this issue to @KaparthyReddy under GSSoC'26.

Please keep the PR focused on:

* Validating supported audio MIME types and/or file extensions
* Rejecting unsupported file types before writing them to disk
* Returning a clear and consistent error response for invalid uploads
* Adding tests covering valid audio uploads and invalid file types

Please pull the latest changes from the main branch before starting work.

Looking forward to the fix.

PR #107, opened 2026-06-05
fix: enforce 50MB file size limit on /record/upload before writing to disk

### LINKED ISSUE #80: [BUG] POST /record/upload has no file size limit — arbitrarily large files written to disk before any validation
In `backend/app/routes/record.py`, the `/record/upload` endpoint writes the uploaded file to disk with no size check:

with open(file_path, "wb") as buffer:
    shutil.copyfileobj(file.file, buffer)

Any authenticated user can upload a file of arbitrary size — gigabytes if they choose — which gets fully written to `data/uploads/` before any validation occurs. This is a denial-of-service vector that can exhaust disk space on the server.

A maximum file size (e.g. 50MB for audio) should be enforced before writing to disk.

KaparthyReddy:
## Description
`/record/upload` used `shutil.copyfileobj` with no size limit:

```python
with open(file_path, "wb") as buffer:
    shutil.copyfileobj(file.file, buffer)
```

Any authenticated user could upload a file of arbitrary size — gigabytes if they chose — which would be fully written to `data/uploads/` before any validation occurred. This is a denial-of-service vector that can exhaust disk space on the server.

Changes made:
- Added a chunked read loop that accumulates upload data in memory using a 64KB read buffer
- Rejects with `HTTP 413 Request Entity Too Large` as soon as `total_bytes` exceeds 50MB — no data is written to disk for oversized files
- The 64KB chunk size avoids blocking the event loop on large uploads while still detecting the limit early
- Also fixed the hardcoded `.wav` extension on saved temp files — now uses the actual extension from `file.filename` so the audio pipeline receives the correct file type
- Works correctly alongside the MIME type and extension validation from PR #81 — size check runs after type validation, before disk write

Fixes #80

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

### DIFF
--- a/backend/app/routes/record.py
+++ b/backend/app/routes/record.py
@@ -49,11 +49,35 @@ async def upload_record(
         # Create a temporary path for the uploaded file
         temp_dir = Path("data/uploads")
         temp_dir.mkdir(parents=True, exist_ok=True)
-        
-        file_path = temp_dir / f"upload_{user_id}_{os.urandom(4).hex()}.wav"
-        
+
+        # Enforce a 50MB file size limit — read the entire upload into
+        # memory in chunks before writing to disk so oversized files are
+        # rejected without filling the disk. shutil.copyfileobj has no
+        # size limit and would write gigabytes before any check ran.
+        MAX_UPLOAD_BYTES = 50 * 1024 * 1024  # 50 MB
+        chunks = []
+        total_bytes = 0
+        chunk_size = 64 * 1024  # 64 KB read buffer
+
+        while True:
+            chunk = await file.read(chunk_size)
+            if not chunk:
+                break
+            total_bytes += len(chunk)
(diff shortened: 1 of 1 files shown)


Chetan0e:
@KaparthyReddy LGTM. Thanks for the contribution.

This change adds an important safeguard to the upload pipeline by enforcing a file size limit before any data is written to disk, helping prevent excessive resource consumption and potential storage abuse.

The implementation is clear, validates uploads early in the request lifecycle, returns an appropriate HTTP 413 response for oversized files, and preserves the correct file extension when saving uploads. The changes are focused, well documented, and address the reported issue effectively.

Approved for merge.

PR #123, opened 2026-06-06
bug#117 fixed

kushagra-1809:
Fix Summary
Bug
When collection.upsert() in ChromaDB partially succeeds then raises an exception (e.g., document-count limits hit mid-batch), the rollback only cleaned up MongoDB via delete_many/delete_one. The vectors already written to ChromaDB were left orphaned — over time these ghosts cause KeyError/None-dereference when vector search returns IDs that no longer exist in MongoDB.

Changes
memory_store.py

store_memories_batch (line ~187–195): Added collection.delete(ids=mem_ids) in the rollback except block, wrapped in a best-effort try/except so that a failed ChromaDB cleanup doesn't mask the original error or prevent the MongoDB rollback.

store_memory (line ~99–106): Applied the same defensive ChromaDB rollback to the single-item path, preventing the same class of orphaned vector even though it's less likely with a single document.

test_store_memories_batch.py

Updated test_rollback_on_chromadb_failure to mock collection.delete and assert it's called with all mem_ids during rollback, so this bug doesn't regress.

### DIFF
--- a/backend/app/services/memory_store.py
+++ b/backend/app/services/memory_store.py
@@ -96,6 +96,15 @@ async def store_memory(
             f"rolling back MongoDB insert: {e}"
         )
 
+        # Rollback ChromaDB: remove the vector if it was partially written
+        try:
+            collection.delete(ids=[mem_id])
+        except Exception as chroma_err:
+            logger.error(
+                f"ChromaDB rollback delete also failed for {mem_id}: "
+                f"{chroma_err}"
+            )
+
         # Rollback Mongo insert to maintain consistency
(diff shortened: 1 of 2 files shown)


Chetan0e:
@kushagra-1809 Merged. 

Thanks for the contribution and for adding regression coverage alongside the fix.

This change improves rollback consistency between MongoDB and ChromaDB by ensuring vector cleanup is attempted when batch or single-memory storage operations fail. The additional tests help prevent orphaned-vector regressions in future changes.

Appreciate the contribution!

PR #127, opened 2026-06-08
feat(auth): per-account login lockout to complement IP rate limiting (#55)

### LINKED ISSUE #55: Brute-force protection: per-account login lockout to complement existing IP rate limiting
## The gap

`backend/app/routes/auth.py` uses slowapi to rate-limit `/login` per IP. That stops a single-machine attacker from hammering one user's credentials, but it doesn't stop:

- Credential stuffing from a botnet — each request comes from a different IP, none of which individually trip the slowapi limit
- Slow brute force from a single IP that stays just under the rate limit
- Distributed attempts coordinated across multiple IPs against one target account

A per-account counter persisted in MongoDB closes that gap. After N failed login attempts on the same username (regardless of source IP), the account is locked for a cooldown window. Resets on a successful login.

## What this would add

A new MongoDB collection `login_attempts` tracking `{username, failures, last_failed_at, locked_until}`. The `/login` route changes:

1. Before checking the password, read the record. If `locked_until > now()`, return 429 with seconds remaining.
2. On password mismatch, increment `failures` and set `last_failed_at`. If `failures >= LOGIN_MAX_FAILURES`, set `locked_until = now() + LOGIN_LOCKOUT_MINUTES`.
3. On successful auth, delete the record.

Defaults: 5 failures, 15-minute lockout. Both configurable via `.env`.

## Why this matters

The README's Security Features list claims JWT with refresh rotation, bcrypt hashing, and input validation, but doesn't address brute force beyond what slowapi does at the IP layer. For a personal memory system holding sensitive recorded conversations, that's a meaningful gap — the threat model is exactly the kind of attacker who'd patiently slow-brute from a distributed source.

## Rough implementation

- Lockout logic in `backend/app/services/auth.py` (read/increment/reset)
- Integration point in the existing `/login` handler in `backend/app/routes/auth.py`
- Config knobs `LOGIN_MAX_FAILURES` and `LOGIN_LOCKOUT_MINUTES` in `.env.example` and `backend/app/config.py`
- Log lockout events through the existing logging in `backend/app/core/logging_config.py`
- Tests: increments on failure, resets on success, lockout returns 429, expires after the configured window

Happy to take this on under GSSoC'26.

Yuthika10:
Closes #55.

Adds a per-account login lockout that complements the existing slowapi IP rate limiting. After N failed login attempts on a username (regardless of source IP), the account is locked for a cooldown window and `/login` returns 429 with a `Retry-After` header. The counter resets on a successful login.

**What changed**

- `backend/app/config.py` — two new settings: `login_max_failures` (default 5) and `login_lockout_minutes` (default 15), reachable via `.env`.
- `backend/app/services/auth.py` - three async helpers next to the existing user logic: `get_lockout_seconds_remaining`, `register_failed_login`, `reset_login_attempts`. Backed by a new `login_attempts` MongoDB collection (`{username, failures, last_failed_at, locked_until}`). Matches the file's `datetime.utcnow()` and `username.lower().strip()` conventions.
- `backend/app/routes/auth.py` - `/login` now checks lockout first, then authenticates, registers a failure before raising 401, and resets attempts before issuing tokens on success. `signup`, `refresh`, and `logout` are unchanged.
- `backend/app/services/database.py` - `login_attempts` indexes: unique on `username` and a TTL on `locked_until`, mirroring the existing `blacklisted_tokens` pattern.
- `.env.example` - `LOGIN_MAX_FAILURES` and `LOGIN_LOCKOUT_MINUTES` documented.
- `backend/tests/test_account_lockout.py` - 9 tests: not-locked when no record, active lockout returns remaining seconds, expired lockout returns 0, below-threshold increments without locking, threshold locks the account, first failure on a fresh account increments from 0, success clears the record, username normalization, and an end-to-end route test asserting 429 + `Retry-After` after the threshold.

**Design note worth flagging**

While `locked_until > now()`, login is rejected with 429 regardless of whether the password is correct. The reset only fires on a successful auth that isn't currently locked. This is the secure reading — if a correct password during the lockout window cleared the counter, an attacker who eventually guesses right mid-window would escape the lockout. Happy to switch behavior if a different intent was meant.

**Note on rebase**

Originally branched off `main` before #85 (refresh token rotation) landed. After rebasing onto current `main`, the only conflict was in the `routes/auth.py` import block - kept both upstream's `ALGORITHM` import (used by the new refresh-rotation logic) and the three new lockout names. All 9 tests pass against post-rebase main.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,6 +30,8 @@ DATABASE_NAME=verath
 
 # ── Security (required — min 32 characters) ──────────────────────────────────
 SECRET_KEY=change-me-to-a-long-random-secret-key-at-least-32-chars
+LOGIN_MAX_FAILURES=5
+LOGIN_LOCKOUT_MINUTES=15
 
 # ── Audio processing ───────────────────────────────────────────────────────────
 AUDIO_SAMPLE_RATE=16000
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -45,6 +45,15 @@ class Settings(BaseSettings):
         if v in ("your-super-secret-key", "changeme", "secret"):
             raise ValueError("SECRET_KEY is set to a known insecure default — please change it.")
         return v
+    
+    login_max_failures: int = Field(
+        default=5,
+        description="Failed login attempts before an account is locked"
+    )
+    login_lockout_minutes: int = Field(
+        default=15,
+        description="Account lockout duration in minutes"
+    )
 
     # ── LLM Providers (Groq + Gemini Fallback) ──────────────────────────────────
     groq_api_key: str = Field(default="", description="Groq API Key")
--- a/backend/app/routes/auth.py
+++ b/backend/app/routes/auth.py
@@ -16,6 +16,9 @@ from app.services.auth import (
     verify_refresh_token,
     decode_access_token,
     ALGORITHM,
+    get_lockout_seconds_remaining,
+    register_failed_login,
+    reset_login_attempts,
 )
 from app.config import settings
 from app.services.database import get_db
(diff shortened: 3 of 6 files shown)


Yuthika10:
Apologies for the delay on this - I was out of town for the last 10 days. Just pushed the PR now. Thank you for your patience.

Chetan0e:
@Yuthika10 Thanks for the contribution.

This is a meaningful security improvement that complements the existing IP-based rate limiting with per-account protection against distributed brute-force attacks.

The implementation is configurable, includes appropriate tests, and integrates cleanly with the existing authentication flow. I'll do a final review of the MongoDB TTL index behavior, but overall the approach looks solid.

LGTM overall. Thanks for the detailed implementation and test coverage.

PR #134, opened 2026-06-11
fix(validators): use model_validator so max_length check works (#133)

### LINKED ISSUE #133: [BUG] TextInputValidator and QueryValidator raise AttributeError on every call (cls.max_length in field_validator)
**Describe the bug**

The `TextInputValidator` and `QueryValidator` classes in `backend/app/core/validators.py` raise `AttributeError: max_length` on every validation call, because the `@field_validator` accesses `cls.max_length`.

In Pydantic v2, inside a `@field_validator` classmethod, `cls` is the model class. Field names like `max_length` are not accessible as plain class attributes there — `cls.max_length` raises `AttributeError`. As a result, any non-empty input that reaches the length check fails:

```python
@field_validator('text')
@classmethod
def validate_text(cls, v):
    ...
    if len(v) > cls.max_length:   # AttributeError: max_length
        raise ValueError(...)
```

Both validators have this same pattern (`validators.py` lines 20 and 48). Note: these classes are not currently imported anywhere in the backend, so this is a latent bug — it will surface the moment either validator is wired into a route, at which point all input validation through them breaks.

**To Reproduce**

```python
from pydantic import BaseModel, field_validator

class TextInputValidator(BaseModel):
    text: str
    max_length: int = 10000

    @field_validator('text')
    @classmethod
    def validate_text(cls, v):
        if len(v) > cls.max_length:
            raise ValueError(f'Text exceeds maximum length of {cls.max_length}')
        return v.strip()

TextInputValidator(text="hello world")
# -> AttributeError: max_length
```

(Tested on Pydantic 2.13.)

**Expected behavior**

Validation should compare against the instance's `max_length` and raise a clean `ValueError`/422 when the limit is exceeded — not crash with `AttributeError`. Because `max_length` is itself a field, the length check should run as a `model_validator(mode="after")` (or read the value via `ValidationInfo.data`) rather than as a `field_validator` on `text` that reaches for a sibling field through `cls`.

**Proposed fix (sketch)**

```python
from pydantic import BaseModel, model_validator

class TextInputValidator(BaseModel):
    text: str
    max_length: int = 10000

    @model_validator(mode="after")
    def validate_text(self):
        if not self.text or not self.text.strip():
            raise ValueError("Text cannot be empty")
        if len(self.text) > self.max_length:
            raise ValueError(f"Text exceeds maximum length of {self.max_length}")
        # ... injection-pattern checks ...
        self.text = self.text.strip()
        return self
```

Yuthika10:
## Summary

Fixes #133. `TextInputValidator` and `QueryValidator` in `backend/app/core/validators.py` accessed `cls.max_length` inside a `@field_validator`. In Pydantic v2, `cls` inside a field validator is the model class, and a field name isn't accessible there as a plain attribute - so `cls.max_length` raises `AttributeError` on every non-empty input. Both validators were effectively broken; any input that passed the empty-check crashed instead of validating.

Switched both to `@model_validator(mode="after")` reading `self.max_length` / `self.text` (and `self.query`). This preserves all the original behavior - empty rejection, the injection-pattern checks, whitespace stripping — while actually honoring a custom `max_length`.

## Changes

- `backend/app/core/validators.py`: rewrote `TextInputValidator.validate_text` and `QueryValidator.validate_query` as `model_validator(mode="after")`. Swapped the unused `field_validator` import for `model_validator`.
- `backend/tests/test_validators.py`: new unit tests - valid input is accepted, custom `max_length` is enforced, empty input is rejected, and the three injection patterns (`<script>`, `javascript:`, `on*=`) are rejected. Parallel coverage for `QueryValidator`.
- `backend/app/routes/speaker.py`: fixed an unrelated import error that was blocking the test suite. The file imported `get_current_user_id` but referenced the undefined name `get_current_user` in two route signatures, raising `NameError` at import time and crashing `app.main` (which `conftest.py` imports). Imported the correct name from `app.services.auth`, where both functions are defined. Flagging this explicitly since it's outside the validator scope - happy to split it into its own PR if you'd prefer.

## Testing

`pytest tests/test_validators.py -v` → 10 passed. The validators now construct correctly (verified `TextInputValidator(text="hello world")` returns the value instead of raising), enforce custom limits, and reject empty/dangerous input.

## Notes

The validators aren't imported anywhere in the backend yet, so this was a latent bug - but it would have surfaced the moment either validator was wired into a route. The fix makes them usable.

### DIFF
--- a/backend/app/core/validators.py
+++ b/backend/app/core/validators.py
@@ -1,6 +1,6 @@
 import re
 from typing import Optional, List
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, model_validator 
 from fastapi import HTTPException, status
 from pathlib import Path
 from app.core.logging_config import logger
@@ -10,45 +10,45 @@ class TextInputValidator(BaseModel):
     """Validator for text inputs to prevent injection attacks."""
     text: str
     max_length: int = 10000
-    
-    @field_validator('text')
-    @classmethod
-    def validate_text(cls, v):
-        if not v or not v.strip():
+
+    @model_validator(mode="after")
+    def validate_text(self):
+        if not self.text or not self.text.strip():
             raise ValueError("Text cannot be empty")
-        
-        if len(v) > cls.max_length:
-            raise ValueError(f"Text exceeds maximum length of {cls.max_length}")
-        
+
+        if len(self.text) > self.max_length:
+            raise ValueError(f"Text exceeds maximum length of {self.max_length}")
+
         # Check for potential injection patterns
         dangerous_patterns = [
             r'<script.*?>.*?</script>',
             r'javascript:',
             r'on\w+\s*=',
         ]
-        
+
         for pattern in dangerous_patterns:
-            if re.search(pattern, v, re.IGNORECASE):
+            if re.search(pattern, self.text, re.IGNORECASE):
                 raise ValueError("Text contains potentially dangerous content")
-        
-        return v.strip()
+
+        self.text = self.text.strip()
+        return self
 
 
 class QueryValidator(BaseModel):
     """Validator for query inputs."""
     query: str
     max_length: int = 500
-    
-    @field_validator('query')
-    @classmethod
-    def validate_query(cls, v):
-        if not v or not v.strip():
+
+    @model_validator(mode="after")
+    def validate_query(self):
+        if not self.query or not self.query.strip():
             raise ValueError("Query cannot be empty")
-        
-        if len(v) > cls.max_length:
-            raise ValueError(f"Query exceeds maximum length of {cls.max_length}")
-        
-        return v.strip()
+
+        if len(self.query) > self.max_length:
+            raise ValueError(f"Query exceeds maximum length of {self.max_length}")
+
+        self.query = self.query.strip()
+        return self
 
 
 def sanitize_filename(filename: str) -> str:
--- a/backend/app/routes/speaker.py
+++ b/backend/app/routes/speaker.py
@@ -2,7 +2,7 @@ from fastapi import APIRouter, Depends
 from app.models.schema import VoiceTrainRequest
 from app.services.gemini_embedding import get_embedding
 from app.services.speaker_training import add_voice, get_voice_profiles
-from ..services.auth import get_current_user_id
+from ..services.auth import get_current_user_id, get_current_user
 
 router = APIRouter()
 
--- a/backend/tests/test_validators.py
+++ b/backend/tests/test_validators.py
@@ -1,0 +1,46 @@
+import pytest
+from pydantic import ValidationError
+
+from app.core.validators import TextInputValidator, QueryValidator
+
+
+class TestTextInputValidator:
+    def test_accepts_valid_text(self):
+        # Regression: this previously raised AttributeError: max_length
+        v = TextInputValidator(text="hello world")
+        assert v.text == "hello world"
+
+    def test_strips_whitespace(self):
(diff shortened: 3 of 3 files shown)


Yuthika10:
Hey @Chetan0e 
Heads-up on the CI `test` job: it runs the full suite, and 37 tests fail because the runner has no MongoDB (ServerSelectionTimeoutError on localhost:27017), no live server (httpx ConnectError in test_e2e), and no Gemini/Groq API keys. A few others (test_real_world intent/correction/temporal cases) are pre-existing assertion mismatches in the NLP heuristics, unrelated to this change.

These all fail on a clean checkout of main too - this PR doesn't touch any of that code. What this PR changes (app/core/validators.py) is covered by tests/test_validators.py, which passes (10/10). The speaker.py import fix was needed just so the suite could be collected at all (it was crashing app.main at import).

Happy to help wire up CI services (a MongoDB service container, mocked API keys) as a separate PR if that'd be useful, but didn't want to fold unrelated CI infrastructure work into a validator bug fix.

Chetan0e:
@Yuthika10 Thanks for the detailed investigation and for including regression tests.

The validator root cause and the switch to `model_validator(mode="after")` look correct, and the added test coverage is helpful for preventing regressions.

I do have one concern before merging: this PR currently includes both the validator fix (#133) and an unrelated import fix in `routes/speaker.py`.

The import issue appears valid, but it is outside the scope of the validator bug. If possible, please either:

* move the speaker import change into a separate PR, or
* confirm that the change is strictly required for test collection and contains no behavioral changes beyond fixing the missing import.

Once that's clarified, I'll proceed with the final review.

Yuthika10:
Thank you @Chetan0e! On the speaker.py change - it's strictly required for test collection and has no behavioral change beyond fixing the missing import.

`conftest.py` imports `app.main`, which imports `app.routes.speaker`. That module imported `get_current_user_id` but referenced the undefined name `get_current_user` in two route signatures (`train_voice`, and the other speaker route), so it raised `NameError` at import time and crashed `app.main` — which blocked the *entire* test suite from even collecting, including tests/test_validators.py. The one-line fix imports the correct name from `app.services.auth` (where both `get_current_user_id` and `get_current_user` are defined). No logic, signatures, or behavior changed — it just makes the previously-undefined name resolve.

Would be happy to split it into its own PR if you'd still prefer that for a clean history - just let me know and I'll pull it out and rebase this PR to validators-only. Whatever's easier for your review.

Chetan0e:
@Yuthika10 Thanks for the clarification and the detailed investigation.

I reviewed the changes and confirmed that the `speaker.py` modification is strictly required for test collection. Since `conftest.py` imports `app.main`, which imports `app.routes.speaker`, the missing `get_current_user` import causes a `NameError` during module import and prevents the test suite from being collected. The fix only resolves that missing reference and does not introduce any behavioral changes.

Regarding the validator fix, the root cause has been correctly identified. Accessing `cls.max_length` inside a Pydantic v2 `field_validator` is invalid in this case, causing validation to fail for non-empty inputs. Migrating to `model_validator(mode="after")` and using instance attributes (`self.max_length`, `self.text`, and `self.query`) is the correct approach.

The added regression tests provide solid coverage for:

* Valid input acceptance
* Whitespace trimming
* Custom `max_length` enforcement
* Empty input rejection
* Injection-pattern detection

The implementation remains focused on Issue #133 and includes appropriate test coverage to prevent future regressions.

Approved. Thank you for the thorough investigation, clear documentation, and comprehensive tests.

PR #136, opened 2026-06-11
fix: clear lru_cache on transcription error to allow model reload

### LINKED ISSUE #110: [Bug] transcription.py uses lru_cache on get_model() — Whisper model is never reloaded if it silently becomes stale or corrupted after initial load
### Description
`backend/app/services/transcription.py` caches the Whisper model with `@lru_cache(maxsize=1)`:

```python
@lru_cache(maxsize=1)
def get_model() -> WhisperModel:
    ...
    return WhisperModel(settings.whisper_model, ...)
```

`lru_cache` caches the return value permanently for the lifetime of the process with no TTL or invalidation mechanism. If the Whisper model becomes corrupted, runs out of GPU memory mid-session, or if `settings.whisper_model` changes (e.g. via environment variable update without restart), the cached stale model is used indefinitely with no way to force a reload short of restarting the entire server.

Additionally, if `get_model()` raises on first call (e.g. model download fails), `lru_cache` does **not** cache exceptions — the next call retries. But if it succeeds and then the model becomes unusable, there is no recovery path.

### Steps to Reproduce
1. Load the model successfully
2. Exhaust GPU memory causing the model to enter an error state
3. All subsequent transcription calls use the same broken model instance

### Expected Behavior
The model cache should have a way to be invalidated (e.g. `get_model.cache_clear()`) which should be called in the `except` block of `transcribe()` when a transcription error occurs, so the next call attempts to reload the model.

### Screenshots
N/A — silent model degradation.

### Environment
- OS: Any
- Python: >=3.11, CUDA optional
- Version: Current upstream

KaparthyReddy:
## Description
`transcription.py` cached the Whisper model with `@lru_cache(maxsize=1)` with no invalidation mechanism:

```python
@lru_cache(maxsize=1)
def get_model() -> WhisperModel:
    ...
    return WhisperModel(settings.whisper_model, ...)
```

If the Whisper model entered an error state after initial load (e.g. GPU OOM, model corruption, exhausted memory mid-session), all subsequent `transcribe()` calls reused the same broken instance indefinitely for the lifetime of the process. There was no recovery path short of restarting the entire server.

Changes made:
- Added `get_model.cache_clear()` in the `except` block of `transcribe()` — evicts the cached model instance when a transcription error occurs
- The next call to `transcribe()` will call `get_model()` again and reload a fresh model instance
- Added `logger.info` to make the cache clear visible in server logs for observability
- The `lru_cache` is preserved for normal operation — the model is still cached and reused across successful transcription calls

Fixes #110

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

### DIFF
--- a/backend/app/services/transcription.py
+++ b/backend/app/services/transcription.py
@@ -36,4 +36,10 @@ def transcribe(audio_path: str) -> str:
         return text
     except Exception as e:
         logger.error(f"Transcription error: {e}")
+        # Clear the lru_cache so the next call attempts to reload the model
+        # rather than reusing a potentially corrupted or stale instance.
+        # Without this, a model that enters an error state (e.g. GPU OOM)
+        # is used indefinitely for the lifetime of the process.
+        get_model.cache_clear()
+        logger.info("Whisper model cache cleared — will reload on next transcription attempt.")
         raise


Chetan0e (changes requested):
Nice fix — this addresses the stale cached Whisper model issue and provides a recovery path without requiring server restart.

I like that the change is minimal and only affects the failure path.

One concern: cache is cleared for every transcription exception, including non-model-related failures (e.g. invalid audio file, path errors). That may trigger unnecessary model reloads.

It may be worth limiting cache invalidation to model/inference-related exceptions only.

Also, I’d strongly recommend adding tests to verify:

cache is cleared on transcription failure
cache is not cleared on successful transcription

Otherwise, LGTM.

PR #145, opened 2026-06-14
fix: reset refreshing state on timeline fetch failure

### LINKED ISSUE #139: [Bug] Timeline pull-to-refresh remains stuck when timeline fetch request fails
### Description
The Timeline screen's pull-to-refresh handler does not reset the refreshing state when the timeline API request throws an exception.

Currently, `setRefreshing(true)` is called before the request starts and `setRefreshing(false)` only after a successful response. If the request fails due to a network issue, server error, or timeout, the function exits before resetting the state.

As a result, the refresh spinner can remain visible indefinitely, giving users the impression that the app is stuck or still loading.

**File:** `mobile/screens/TimelineScreen.js`, lines 33–38

### Steps to Reproduce
This can be reproduced by:
1. Open the Timeline screen.
2. Disable the internet connection (or force the timeline API request to fail).
3. Perform a pull-to-refresh action.
4. Observe the refresh indicator.

### Current Behavior
- The refresh spinner remains active indefinitely after the request fails.
- Users receive no indication that the refresh process has completed or failed.

### Expected Behavior
- The refresh state should always be cleared regardless of whether the API request succeeds or fails.
- The loading indicator should stop once the refresh attempt finishes.
- An error message should inform the user that the refresh failed, consistent with the existing error-handling pattern in `fetchTimeline`.

### Suggested Fix
Wrap the refresh logic in a `try...catch...finally` block, reusing the existing `error` state:

\`\`\`javascript
const onRefresh = async () => {
  setRefreshing(true);
  try {
    const data = await getTimeline();
    setEvents(data.timeline || []);
  } catch (e) {
    setError('Failed to refresh timeline');
  } finally {
    setRefreshing(false);
  }
};
\`\`\`

### Impact
This issue affects user experience on unstable networks and can make the app appear frozen or continuously loading after a failed refresh attempt.

Raising as part of GSSoC 2026 contribution.

VanshikaAsthana:
## Summary
Fixed the Timeline screen's pull-to-refresh handler so the refreshing 
state is always reset, even when the API request fails.

`onRefresh` previously called `setRefreshing(false)` only after a 
successful response. If `getTimeline()` threw an exception, the spinner 
remained active indefinitely. Wrapped the logic in `try...catch...finally` 
so `setRefreshing(false)` always executes, and added error handling via 
the existing `setError` state consistent with `fetchTimeline`.

## Test plan
- [ ] Simulated failed timeline request (network disabled) — spinner stops correctly
- [ ] Successful refresh behavior unchanged

Fixes #139

### DIFF
--- a/mobile/screens/TimelineScreen.js
+++ b/mobile/screens/TimelineScreen.js
@@ -31,11 +31,16 @@ export default function TimelineScreen() {
   };
 
   const onRefresh = async () => {
-    setRefreshing(true);
+  setRefreshing(true);
+  try {
     const data = await getTimeline();
     setEvents(data.timeline || []);
+  } catch (e) {
+    setError('Failed to refresh timeline');
+  } finally {
     setRefreshing(false);
-  };
+  }
+};
 
   const playAudio = async (audioFile, eventId) => {
     try {


VanshikaAsthana:
Hi @Chetan0e, let me know if you'd like any changes to this PR.

Chetan0e (approved):
@VanshikaAsthana Thanks for the clean fix.

I reviewed the PR and the change correctly addresses the issue where the refresh spinner could remain active indefinitely if `getTimeline()` failed.

Using `try...catch...finally` here is the right approach:
- successful refresh behavior remains unchanged
- failed requests now surface an error
- `setRefreshing(false)` is guaranteed to execute

The scope is minimal and focused with no unrelated changes.

One small non-blocking suggestion for future improvement: consider surfacing the actual error message (`e?.message`) instead of a generic string where appropriate.

Approved from my side. Nice work.

PR #171, opened 2026-06-18
fix(task-queue): add asyncio.Lock to prevent concurrent double-initialization race

### LINKED ISSUE #144: Bug: `TaskQueue._ensure_initialized` has a non-atomic double-initialization race under concurrent async tasks
### Summary
`TaskQueue._ensure_initialized` in `backend/app/workers/task_queue.py` (lines 59–81) uses a plain boolean `self._initialized` as a gate, with no `asyncio.Lock`. Under Python's cooperative multitasking, two coroutines can both read `self._initialized = False`, both enter the initialization body, and concurrently call `create_index` on the same collections. This causes duplicate index creation attempts and, more critically, can leave `self._initialized = True` set by the first coroutine while the second is still mid-initialization and has not yet completed all index creations.

### Reproduction
1. Start the application with several concurrent requests arriving before any task is processed (e.g., multiple `/record` or pipeline calls at startup).
2. Multiple coroutines call `task_queue.enqueue()` or `task_queue.dequeue()` simultaneously.
3. Each hits `_ensure_initialized`; `self._initialized` is False for all of them at the first `await` yield point.
4. All proceed into the initialization block concurrently.

### Root Cause
```python
# task_queue.py lines 59–78
async def _ensure_initialized(self):
    if self._initialized:       # ← not protected by a lock
        return
    db = get_db()
    ...
    await db[COLLECTION_NAME].create_index(...)   # ← yields here; others enter concurrently
    ...
    self._initialized = True    # ← set by first finisher; others may still be running
```
`asyncio` tasks interleave at every `await` point. A lock (`asyncio.Lock`) is required to make the check-and-set atomic.

### Affected files
- `backend/app/workers/task_queue.py` — `TaskQueue._ensure_initialized()`, lines 59–81
- `backend/app/workers/task_queue.py` — `TaskQueue.__init__`, line 55–57 (no lock instantiated)

### Impact
Under concurrent startup load, index creation may be attempted multiple times in parallel, potentially causing `OperationFailure` exceptions that propagate to callers as a `RuntimeError`, dropping queued tasks. In the worst case, `self._initialized = True` is set while a sibling coroutine is still creating indexes, causing subsequent callers to skip initialization and operate against a partially-initialized collection.

hariom888:
## Summary

Fixes #144.

## Root Cause

`_ensure_initialized` used a plain `bool` (`self._initialized`) as its only guard. Since Python's asyncio scheduler yields at every `await`, multiple coroutines could all read `self._initialized = False` before any of them completed index creation, then proceed to call `create_index` concurrently. The first coroutine to finish set `self._initialized = True` while siblings were still awaiting their own `create_index` calls — leaving the queue in a partially initialized state visible to subsequent callers.

## Fix

Two targeted changes in `backend/app/workers/task_queue.py`:

1. **`__init__`** — instantiate `self._init_lock = asyncio.Lock()`.
2. **`_ensure_initialized`** — wrap the initialization body with `async with self._init_lock` and add a second `if self._initialized: return` check *inside* the lock (standard async double-checked locking pattern).

```python
async def _ensure_initialized(self):
    if self._initialized:          # fast path — no lock overhead after first init
        return
    async with self._init_lock:
        if self._initialized:      # re-check: another coroutine may have finished
            return
        # ... index creation ...
        self._initialized = True
```

This guarantees:
- Only one coroutine executes the initialization body to completion.
- All other concurrent callers either exit on the outer fast-path or wait on the lock and exit on the inner re-check.
- `self._initialized` is never set `True` before all indexes are created.

## Tests added — `backend/tests/test_task_queue_init_race.py`

| Test | What it verifies |
|---|---|
| `test_create_index_called_once_per_index_under_concurrent_callers` | 10 concurrent callers → `create_index` called exactly 7 times total |
| `test_initialized_flag_is_true_after_concurrent_callers` | `_initialized` is `True` after gather |
| `test_no_runtime_error_raised_under_concurrent_callers` | No exceptions propagate |
| `test_second_call_is_noop_after_initialization` | Subsequent calls are no-ops |
| `test_raises_runtime_error_when_db_unavailable` | `RuntimeError` propagates when `get_db()` returns `None` |
| `test_initialized_remains_false_on_index_creation_failure` | `_initialized` stays `False` on partial failure |
| `test_enqueue_triggers_initialization_exactly_once` | `enqueue()` concurrent callers initialize once |

All mocks use `asyncio.sleep(0)` inside `create_index` to force cooperative interleaving and expose the race reliably.

## Scope

Changes are limited to `__init__` and `_ensure_initialized` only. No queue logic, retry behaviour, or DB access patterns are altered.

## Checklist
- [x] `asyncio.Lock` added to `__init__`
- [x] Double-checked locking pattern in `_ensure_initialized`
- [x] `_initialized = True` only set after all indexes are created
- [x] Concurrent startup tests covering 10+ simultaneous callers
- [x] No unrelated changes

### DIFF
--- a/backend/app/workers/task_queue.py
+++ b/backend/app/workers/task_queue.py
@@ -54,32 +54,42 @@ class TaskQueue:
     
     def __init__(self):
         self._initialized = False
+        self._init_lock = asyncio.Lock()
         self.worker_id = str(uuid.uuid4())
     
     async def _ensure_initialized(self):
-        """Ensure MongoDB connection and indexes are initialized."""
-        if self._initialized:
+        """Ensure MongoDB connection and indexes are initialized.
+
+        Uses an asyncio.Lock with a double-checked locking pattern to guarantee
+        that index creation runs exactly once even when multiple coroutines call
+        this method concurrently at startup.
+        """
+        if self._initialized:          # fast path — no lock needed after first init
             return
-        
-        db = get_db()
-        if db is None:
-            raise RuntimeError("Database not available for task queue")
-        
-        # Create indexes for efficient querying
-        try:
-            await db[self.COLLECTION_NAME].create_index([("task_id", 1)], unique=True)
-            await db[self.COLLECTION_NAME].create_index([("status", 1), ("next_retry_at", 1)])
-            await db[self.COLLECTION_NAME].create_index([("user_id", 1)])
-            await db[self.COLLECTION_NAME].create_index([("heartbeat_at", 1)])
-            await db[self.DEAD_LETTER_COLLECTION].create_index([("task_id", 1)], unique=True)
-            await db[self.DEAD_LETTER_COLLECTION].create_index([("user_id", 1)])
-            await db[self.DEAD_LETTER_COLLECTION].create_index([("failed_at", -1)])
-            
-            self._initialized = True
-            logger.info("Task queue initialized with indexes")
-        except Exception as e:
-            logger.error(f"Failed to initialize task queue indexes: {e}", exc_info=True)
-            raise
+
+        async with self._init_lock:
+            if self._initialized:      # re-check: another coroutine may have finished
+                return                 # while we were waiting for the lock
+
+            db = get_db()
+            if db is None:
+                raise RuntimeError("Database not available for task queue")
+
+            # Create indexes for efficient querying
+            try:
+                await db[self.COLLECTION_NAME].create_index([("task_id", 1)], unique=True)
+                await db[self.COLLECTION_NAME].create_index([("status", 1), ("next_retry_at", 1)])
+                await db[self.COLLECTION_NAME].create_index([("user_id", 1)])
+                await db[self.COLLECTION_NAME].create_index([("heartbeat_at", 1)])
+                await db[self.DEAD_LETTER_COLLECTION].create_index([("task_id", 1)], unique=True)
+                await db[self.DEAD_LETTER_COLLECTION].create_index([("user_id", 1)])
+                await db[self.DEAD_LETTER_COLLECTION].create_index([("failed_at", -1)])
+
+                self._initialized = True
+                logger.info("Task queue initialized with indexes")
+            except Exception as e:
+                logger.error(f"Failed to initialize task queue indexes: {e}", exc_info=True)
+                raise
     
     async def enqueue(self, task: Task) -> bool:
         """Enqueue a task for processing."""
--- a/backend/tests/test_task_queue_init_race.py
+++ b/backend/tests/test_task_queue_init_race.py
@@ -1,0 +1,139 @@
+"""
+Tests for TaskQueue._ensure_initialized concurrency safety.
+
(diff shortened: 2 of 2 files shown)


hariom888:
Hi @Chetan0e,

I've reviewed all CI failures and confirmed that **none are caused by this PR**.

---

### Trivy Security Scan

The **Trivy Security Scan** failure is unrelated to this PR. The CI error is:
```
Unable to resolve action `aquasecurity/trivy-action@0.30.0`, unable to find version `0.30.0`
```
This fails at the "Set up job" step before any code is scanned — the version `0.30.0` of `aquasecurity/trivy-action` no longer resolves. All other security checks (CodeQL, Python Dependency Audit, NPM Dependency Audit) passed successfully.

The workflow file at `.github/workflows/` will need to be updated to a valid Trivy action version (e.g. `aquasecurity/trivy-action@0.31.0` or `@master`). This is a pre-existing infrastructure issue affecting all PRs, not introduced by this change.

---

### CI / Test (41 failures)

All 41 test failures are also pre-existing and fall into these categories:

| Failure category | Root cause |
|---|---|
| `test_e2e.*` (11 tests) | `httpx.ConnectError` — no live server running in CI |
| `test_system.*` (4 tests) | `status=running` vs `ok`, rate limiter bleeding across tests |
| `test_auth.*` (4 tests) | Rate limiter state leaking between tests (429 responses) |
| `test_health.*` (3 tests) | `mock_get_db` is a coroutine not being awaited — mock bug |
| `test_privacy.*` (5 tests) | `NoneType` subscript — DB mock returning None |
| `test_reminders.*` (2 tests) | `'alert_123'` is not a valid ObjectId — test uses fake non-ObjectId |
| `test_real_world.*` (7 tests) | NLP intent/correction logic mismatches, missing `GOOGLE_API_KEY` |
| `test_memory_pipeline.*` (1 test) | `GOOGLE_API_KEY` not set in CI environment |
| `test_account_lockout.*` (1 test) | Missing `Retry-After` header in rate limit response |

My new tests (`tests/test_task_queue_init_race.py`) do not appear in the failed list. The 105 passing tests confirm the existing suite is unaffected by this change.

Chetan0e (approved):
Merged — thanks for the thorough investigation and clean implementation.

This PR resolves the async double-initialization race in `TaskQueue._ensure_initialized()` by introducing an `asyncio.Lock` with a proper double-checked locking pattern.

What I liked:

* Clear root-cause analysis of the race condition under concurrent async callers
* Minimal and focused changes limited to initialization logic
* Correct use of `asyncio.Lock` to ensure only one coroutine performs index creation
* Strong concurrency-focused test coverage covering initialization success, failure, no-op behavior, and concurrent enqueue scenarios

I also appreciate the detailed CI analysis clarifying that the current failures are unrelated to this PR and stem from existing infrastructure/test-environment issues.

This improves startup reliability and prevents inconsistent task queue initialization under concurrent load.

Great work on both the implementation and the supporting tests

PR #173, opened 2026-06-20
fix(auth): route /logout through get_current_user and handle duplicate blacklist insert (#154)

### LINKED ISSUE #154: Security: `/logout` bypasses auth middleware, allowing blacklisted tokens to trigger unhandled DuplicateKeyError (HTTP 500) and rendering token revocation bypassable
Noticed two compounding security bugs in the logout flow while reading through `backend/app/routes/auth.py` and `backend/app/services/auth.py`.

---

## Bug 1 — `/logout` bypasses the standard auth middleware

The `/logout` endpoint manually parses the `Authorization` header and calls `decode_access_token()` instead of using `Depends(get_current_user)`:

```python
# routes/auth.py
@router.post("/logout")
async def logout(request: Request):       # ← no Depends(get_current_user)
    auth_header = request.headers.get("Authorization")
    ...
    token = auth_header.split(" ")[1]
    payload = decode_access_token(token)  # ← only calls jwt.decode(), never checks blacklist
```

`decode_access_token()` in `services/auth.py` is a plain `jwt.decode()` with no blacklist lookup:

```python
def decode_access_token(token: str) -> Optional[Dict]:
    try:
        payload = jwt.decode(token, settings.secret_key, algorithms=[ALGORITHM])
        return payload   # ← blacklist never consulted
    except JWTError:
        return None
```

This means a token that has already been blacklisted — after a previous logout or a refresh rotation — is treated as valid by the logout endpoint. Every other protected route uses `Depends(get_current_user)` which does check the blacklist, so the logout route is the only one inconsistently handling this.

---

## Bug 2 — Re-submitting a blacklisted token to `/logout` crashes the server with HTTP 500

`database.py` creates a unique index on `blacklisted_tokens.jti`:

```python
await db["blacklisted_tokens"].create_index("jti", unique=True)
```

When a blacklisted token is submitted to `/logout` again, the endpoint reaches:

```python
await db["blacklisted_tokens"].insert_one({"jti": jti, ...})
```

This raises `pymongo.errors.DuplicateKeyError` which is not caught anywhere in the route, resulting in an unhandled HTTP 500.

---

## Reproduction steps

1. Log in → receive token A
2. Call `POST /logout` with token A → A is blacklisted, logout succeeds
3. Call `POST /logout` again with the same token A
4. `decode_access_token` decodes it successfully (no blacklist check)
5. `insert_one` raises `DuplicateKeyError` → **HTTP 500**

Any party in possession of a user's old access token (from browser history, a log file, or network interception) can repeatedly trigger this crash, constituting a targeted DoS against the auth service.

---

## Root cause summary

| File | Location | Issue |
|---|---|---|
| `routes/auth.py` | `async def logout(request: Request)` | Missing `Depends(get_current_user)` — blacklist never checked |
| `routes/auth.py` | `db["blacklisted_tokens"].insert_one(...)` | No `try/except` — `DuplicateKeyError` crashes the server |
| `services/auth.py` | `decode_access_token()` | Only used by logout; skips blacklist check unlike `verify_access_token` |

---

## Suggested fix

Replace manual header parsing with the standard dependency and handle the duplicate case:

```python
from pymongo.errors import DuplicateKeyError

@router.post("/logout")
async def logout(
    request: Request,
    current_user: dict = Depends(get_current_user),  # validates + checks blacklist
    creds: HTTPAuthorizationCredentials = Depends(HTTPBearer()),
):
    payload = decode_access_token(creds.credentials)
    jti = payload.get("jti")
    exp = payload.get("exp")

    if jti and exp:
        db = get_db()
        try:
            await db["blacklisted_tokens"].insert_one({
                "jti": jti,
                "exp": datetime.utcfromtimestamp(exp),  # utc, not local time
                "blacklisted_at": datetime.utcnow(),
                "username": current_user["username"],
            })
        except DuplicateKeyError:
            pass  # already blacklisted — treat as successful logout

    await _log_auth_event(current_user["username"], request.client.host, "logout", True)
    return {"message": "Logged out successfully"}
```

Can put up a PR with the fix and tests covering both the re-logout crash and the blacklist bypass if that looks right.

Yuthika10:
## Summary

Closes #154.

Fixes two compounding security bugs in the logout flow (both verified against current `main`):

**Bug 1 — `/logout` bypassed the blacklist check.** The endpoint parsed the `Authorization` header manually and called `decode_access_token()` (plain `jwt.decode`, no blacklist lookup) instead of the `get_current_user` dependency that every other protected route uses. `get_current_user` → `verify_access_token` checks `blacklisted_tokens`; `decode_access_token` does not. So an already-blacklisted token (from a previous logout or refresh rotation) was treated as valid at `/logout`.

**Bug 2 — re-submitting a blacklisted token returned HTTP 500.** `blacklisted_tokens.jti` has a unique index. A second logout with the same token reached `insert_one` and raised an uncaught `DuplicateKeyError` → unhandled 500. Anyone holding an old access token could repeatedly trigger this — a targeted DoS on the auth service.

## Changes

- `/logout` now depends on `get_current_user`, so it goes through `verify_access_token` and rejects blacklisted tokens (401) — consistent with every other protected route. A blacklisted token never reaches the insert.
- Kept the bearer credentials (via `HTTPBearer`) to extract `jti`/`exp` for blacklisting, since `get_current_user` returns only the username.
- Wrapped `insert_one` in `try/except DuplicateKeyError`, so a repeat/raced logout is idempotent (200) rather than a 500 — defense-in-depth even though Bug 1's fix already prevents a blacklisted token from reaching it.

I kept `datetime.fromtimestamp(exp)` as-is rather than changing it — the local-vs-UTC question is a separate concern and out of scope for this fix.

## Testing

Added two tests to `backend/tests/test_auth.py`:
- `test_logout_rejects_already_blacklisted_token` — a blacklisted token is rejected at `/logout` with 401 (fails on the old code, which returned 200 — the bypass proof).
- `test_relogout_does_not_500_on_duplicate` — when `insert_one` raises `DuplicateKeyError`, logout returns 200 instead of crashing.

Both pass, along with the existing logout/blacklist tests (7 passed, no regressions):

```text
tests/test_auth.py::TestAuth::test_logout_still_invalidates_access_token PASSED
tests/test_auth.py::TestAuth::test_logout_rejects_already_blacklisted_token PASSED
tests/test_auth.py::TestAuth::test_relogout_does_not_500_on_duplicate PASSED
... (7 passed, 7 deselected)
```

Note: the other tests in `test_auth.py` (signup/login/refresh) require a live MongoDB at `localhost:27017` and fail locally without one - this is pre-existing and unrelated to this change (those tests don't exercise `/logout`).

### DIFF
--- a/backend/app/routes/auth.py
+++ b/backend/app/routes/auth.py
@@ -1,5 +1,7 @@
 import logging
-from fastapi import APIRouter, HTTPException, status, Request
+from fastapi import APIRouter, HTTPException, status, Request, Depends
+from fastapi.security import HTTPBearer, HTTPAuthorizationCredentials
+from pymongo.errors import DuplicateKeyError
 from pydantic import BaseModel
 from slowapi import Limiter, _rate_limit_exceeded_handler
 from slowapi.util import get_remote_address
@@ -15,6 +17,7 @@ from app.services.auth import (
     create_refresh_token,
     verify_refresh_token,
     decode_access_token,
+    get_current_user,
     ALGORITHM,
     get_lockout_seconds_remaining,
     register_failed_login,
@@ -25,6 +28,8 @@ from app.services.database import get_db
 
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/auth", tags=["auth"])
+
+_bearer = HTTPBearer()
 
 # Rate limiting configuration
 limiter = Limiter(key_func=get_remote_address)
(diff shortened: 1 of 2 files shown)


Yuthika10:
Hi @Chetan0e! This a PR for issue #154. Please go through the PR would love your feedback. 
Also, the two tests added here for this fix pass in CI:
- test_logout_rejects_already_blacklisted_token
- test_relogout_does_not_500_on_duplicate

The other failures in the run are pre-existing and unrelated to this change - they're in test_e2e/test_privacy/test_real_world/test_system and fail because the CI environment doesn't provide the external services those tests need (MongoDB connection, Groq/Gemini API keys, and rate-limit state), not because of anything in the logout flow. This change only touches /logout and its tests. Happy to look at the broader CI/test-environment setup separately if that'd be useful, but it's out of scope here.

Chetan0e (approved):
@Yuthika10 Thanks for the thorough fix and clear test coverage.

I reviewed the changes and the implementation correctly addresses both issues from #154:

* `/logout` now routes through `get_current_user`, ensuring blacklisted tokens are rejected consistently with other protected routes.
* `DuplicateKeyError` is handled during blacklist insertion, preventing repeated logout attempts from causing HTTP 500.
* Added regression tests cover both the blacklist bypass and duplicate insert cases.

This is a solid security fix and the scope is well contained.

One minor observation (non-blocking):
`get_current_user` already validates and decodes the token internally, but `/logout` currently calls `decode_access_token(creds.credentials)` again to extract `jti`/`exp`. This results in duplicate decoding. In the future we could consider exposing token payload from the auth dependency to avoid that redundancy.

The current implementation is still correct and safe.

Approved from my side.

Issue #174, opened 2026-06-20
[Feature]: Scheduled weekly memory digest — wire the existing summarizer into APScheduler and expose a digest endpoint with configurable delivery cadence


Yuthika10:
## What would you like to add or change?

Add a scheduled weekly digest to Verath — a proactive LLM-generated recap of the user's captured memories, delivered on a fixed cadence via APScheduler and retrievable on demand through a new `/digest` endpoint.

The feature has three parts:

**`app/services/digest.py`** (new file) — generates a per-user summary over a configurable time window and persists it to a `digests` collection:

```python
async def generate_and_store_digest(user_id: str, window_hours: int = 168) -> Dict[str, Any]:
    summary = await generate_period_summary(user_id, hours=window_hours)
    doc = {
        "user_id": user_id,
        "generated_at": datetime.utcnow(),
        "window_hours": window_hours,
        "summary": summary,
    }
    await _digests_col.insert_one(doc)
    return doc

async def run_weekly_digests() -> None:
    """Scheduler job: generate digests for every active user."""
    async for user in users_col.find({}, {"_id": 1}):
        await generate_and_store_digest(str(user["_id"]))
```

**`app/main.py`** — register the job next to the existing reminder check:

```python
from app.services.digest import run_weekly_digests

_scheduler.add_job(
    run_weekly_digests,
    trigger="interval",
    weeks=1,
    id="weekly_digest",
    replace_existing=True,
)
```

**`app/routes/advanced.py`** — two new endpoints:

```python
@router.get("/digest/latest")
async def latest_digest(user_id: str = Depends(get_current_user_id)):
    doc = await get_latest_digest(user_id)
    if not doc:
        raise HTTPException(status_code=404, detail="No digest available yet.")
    return {
        "generated_at": doc["generated_at"].isoformat(),
        "window_hours": doc["window_hours"],
        "summary": doc["summary"],
    }

@router.post("/digest/generate")
async def trigger_digest(
    window_hours: int = Query(168, ge=1, le=720),
    user_id: str = Depends(get_current_user_id),
):
    """Manually trigger a digest for the current user."""
    doc = await generate_and_store_digest(user_id, window_hours=window_hours)
    return {"generated_at": doc["generated_at"].isoformat(), "summary": doc["summary"]}
```

---

## What problem does this solve?

[...]

PR #187, opened 2026-06-25
feat: add weekly digest generation and retrieval (#174)

Yuthika10:
## Summary
 
Implements the weekly digest feature from #174. It reuses the existing `generate_period_summary` from `summarizer.py` to build a per-user summary, stores it, and exposes it through a couple of endpoints plus a scheduled weekly job.
 
## What changed
 
**New service: `app/services/digest.py`**
- `generate_and_store_digest(user_id, window_hours=168)` — calls `generate_period_summary` and stores the result in a `digests` collection.
- `get_latest_digest(user_id)` — returns the most recent digest for a user, or `None`.
- `run_weekly_digests()` — scheduler job that loops over all users, returns a count, and skips a failing user without stopping the rest (same shape as `check_and_fire_reminders`).
I modeled the module-level collection access (`_digests_col`, `_users_col`) on `reminder_service.py` so it stays consistent with the sibling service.
 
**Endpoints (on the advanced router)**
- `GET /digest/latest` — returns the latest stored digest, or 404 if none exists yet.
- `POST /digest/generate` — manually triggers a digest. Takes a `window_hours` query param (default 168 = one week, validated to 1–720).
**Scheduler (`main.py`)**
- Registered `run_weekly_digests` as a weekly APScheduler job, right next to the existing reminder job and following the same `add_job` pattern.
## Tests
 
`tests/test_digest.py` mirrors the mock style in `test_reminders.py`:
- 5 service unit tests: store, empty-window fallback, latest lookup, full user loop, failure isolation.
- 4 endpoint tests: 404 when none, success path, generate, out-of-range validation.
All 9 pass locally.
 
## Notes
- The digest reuses the existing summarizer rather than adding new summarization logic.
- Collection access matches `reminder_service.py` for consistency.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -10,6 +10,7 @@ from app.config import settings
 from app.core.logging_config import setup_logging
 from app.workers.background_worker import start_worker
 from app.services.reminder_service import check_and_fire_reminders
+from app.services.digest import run_weekly_digests
 from app.services.database import get_db
 
 # ── Routers ───────────────────────────────────────────────────────────────────
@@ -141,6 +142,17 @@ async def lifespan(app: FastAPI):
         replace_existing=True,
         misfire_grace_time=60,   # allow 60s late start before skipping
     )
+    
+    # 5. Weekly digest — runs once a week, generates a per-user summary digest
+    _scheduler.add_job(
+        run_weekly_digests,
+        trigger="interval",
+        weeks=1,
+        id="weekly_digest",
+        replace_existing=True,
+        misfire_grace_time=3600,   # allow 1h late start before skipping
+    )
+    
     _scheduler.start()
     logger.info("Reminder scheduler started (interval: 15 min)")
 
--- a/backend/app/routes/advanced.py
+++ b/backend/app/routes/advanced.py
@@ -9,6 +9,7 @@ from app.services.summarizer import generate_daily_summary, extract_key_insights
 from app.services.timeline import get_today_timeline
 from app.services.memory_store import get_memory_stats, all_memories, all_memories_filtered
 from app.services.auth import get_current_user_id
+from app.services.digest import generate_and_store_digest, get_latest_digest
 from app.services.memory_graph import build_memory_graph
 from app.core.logging_config import logger
 from app.core.cache import cached, get_cache_stats, invalidate_cache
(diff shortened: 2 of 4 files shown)


Yuthika10:
Hi @Chetan0e this is PR resolving issue #174 
Heads up on the failing `test` check: all 41 failures are pre-existing on `main` and unrelated to this PR.
 
They're spread across `test_auth`, `test_e2e`, `test_health`, `test_privacy`, `test_real_world`, `test_reminders`, and `test_system`, and the causes are environmental or existing mock issues:
 
- Groq / Gemini API keys not set in CI (the `test_real_world` and `test_memory_pipeline` failures)
- no live server for the e2e tests (`ConnectError` in `test_e2e`)
- existing mock bugs, e.g. `ObjectId('alert_123')` in `test_reminders`, `'coroutine' object has no attribute 'command'` in `test_health`, and rate-limit bleed across the auth/signup tests
The digest feature's own tests all pass: the 9 tests in `tests/test_digest.py` are green, and `app/services/digest.py` shows 100% coverage in the run.
 
If it's useful I can run the suite against `main` without my changes to confirm the same 41 fail there, but I didn't want to touch unrelated test files in this PR.

Yuthika10:
Hi @Chetan0e can you review this PR?

Chetan0e (commented):
@Yuthika10 Thanks for the solid implementation and detailed testing. The feature direction is strong, and I appreciate the effort put into this PR.

I’m not approving yet because I want to resolve a few architectural concerns first:

* shared DB access instead of introducing a new Mongo client in `digest.py`
* digest generation should align with task queue / background worker architecture for scalability
* workload control for inactive users (avoid generating digests for every user by default)

Once these are addressed, I’ll be happy to approve and move toward merge.

Yuthika10:
Thanks @Chetan0e , this all makes sense. I've pushed changes for two of the three and want to check direction on the queue one before doing it.
 
**Shared DB:** Done. Removed the separate `AsyncIOMotorClient` from `digest.py` and switched to the shared `get_db()` from `app.services.database`, fetched at call time (since `get_db()` returns `None` until `connect_to_mongo()` runs at startup). Collections are now `get_db()["digests"]` and `get_db()["memories"]`, same pattern as `task_queue.py`.
 
**Inactive users:** Done. `run_weekly_digests` no longer loops every user. It first queries distinct `user_id`s from `memories` with activity in the window and only digests those. I filtered on the `timestamp` field rather than `created_at` because that's what `get_recent_timeline` (which the summarizer uses) actually filters on, so the active-user set matches what the summary would cover. Worth noting `created_at` is indexed but `timestamp` may not be, so if you'd rather index `timestamp` or switch the activity signal to `created_at`, happy to adjust.
 
**Task queue alignment:** This is the one I want your call on. Looking at the worker code, there seem to be two paths: the legacy in-memory `_queue` in `background_worker.py` does execute real callables, but it's marked deprecated and being phased out per the migration comments. The persistent Mongo-backed queue (`task_queue.py`, via `enqueue_task`) currently only persists task metadata and orchestrates status/retries for the `RECORDING` and `COMPRESSION` task types, and doesn't dispatch arbitrary callables yet.
 
So "route digest through the worker" could mean a few different things depending on where the migration is headed:
 
1. Enqueue per-user digests through the existing `enqueue_task` interface now, so the scheduler tick stays light, and let the heavy lifting move into the consumer as the persistent queue matures.
2. Add a dedicated `DIGEST` `TaskType` and wire it through the persistent consumer (bigger change, touches `task_queue.py` core).
3. Keep it on the scheduler for now but offload via the legacy worker.

I didn't want to pick the architecture for you here since the queue's mid-migration. Which direction do you prefer? I'll wire it up to match.

Chetan0e (approved):
@Yuthika10 Thanks for the quick iteration — the changes look much better now.

I reviewed the latest updates and the main concerns are addressed well:

* Shared DB access is now aligned with the existing architecture via `get_db()`
* Active-user filtering prevents unnecessary digest generation for inactive users
* The service and route structure looks clean
* Test coverage is strong and covers both service + endpoint behavior well

Regarding queue integration: since the task queue migration is still in progress and the persistent queue does not yet support digest-style tasks cleanly, I don’t want to block this PR on that architectural piece.

We can treat queue integration as a follow-up improvement:

* either introducing a dedicated `DIGEST` task type
* or routing digest generation through the persistent worker once that migration is complete

For now, I’m okay with the scheduler-based implementation as long as workload stays limited to active users (which is now addressed).

Approved — nice work on this feature and iteration.

PR #206, opened 2026-06-29
fix(auth): eliminate refresh token rotation race via atomic insert-first blacklisting 

### LINKED ISSUE #113: bug: Refresh Token Rotation Has a Non-Atomic Check-Then-Act Race Condition
**File:** `backend/app/routes/auth.py` (lines 97–130)

### Description

The `/auth/refresh` endpoint verifies the refresh token and then, in a separate non-atomic step, writes the old JTI to the blacklist collection. Between `verify_refresh_token()` (which does a `find_one` on `blacklisted_tokens`) and the subsequent `insert_one`, a concurrent request carrying the same refresh token can pass the `find_one` check and also receive a new token pair. Both requests observe the token as not-yet-blacklisted because neither write has committed yet.

```python
# auth.py:97 — validation reads blacklist
username = await verify_refresh_token(body.refresh_token)

# auth.py:119 — blacklist write happens AFTER validation returns
await db["blacklisted_tokens"].insert_one({...})

# auth.py:133 — new token pair issued
return TokenResponse(access_token=..., refresh_token=...)
```

### Reproduction

1. Authenticate and obtain a refresh token.
2. Fire two simultaneous `POST /auth/refresh` requests with the same token.
3. Both requests complete with HTTP 200 and each returns a *different* new refresh token, meaning the original token was effectively used twice.

### Impact

Refresh token rotation is the primary defense against stolen token replay. This race completely defeats it in multi-instance or async-concurrent deployments.

Srejoye:
## Summary

Fixes #113 — non-atomic check-then-act race condition in `/auth/refresh`.

## Root Cause

`verify_refresh_token()` performed a `find_one` on `blacklisted_tokens` to check whether a JTI had been revoked, then returned control to the route handler, which later called `insert_one` to blacklist the consumed token. These two operations are separate database round-trips with no atomicity between them.

Under concurrent load (or across multiple server instances), two requests carrying the same refresh token could both pass the `find_one` check before either `insert_one` committed — causing one refresh token to be exchanged for two independent token pairs, defeating rotation entirely.

## Atomic Strategy: Insert-First

The fix replaces the read-then-write sequence with an **insert-first** pattern:

1. **Decode only** — the token's signature, expiry, and `type` claim are validated synchronously (no DB read).
2. **Attempt `insert_one` of the old JTI immediately** — before any new tokens are issued.
3. **`DuplicateKeyError` → 401** — the unique index on `blacklisted_tokens.jti` (already in `database.py`) means only one    concurrent caller can win the insert. All others receive `401 Refresh token already used or revoked`.
4. **Issue new token pair** — only reached by the single winner.

The unique index acts as a distributed mutex that works correctly across asyncio concurrency and multiple server instances with no additional infrastructure required.

The redundant `verify_refresh_token()` call in the route (which was itself the source of the race) is removed. Its blacklist check is now superseded by the atomic insert gate.

## Files Changed

| File | Change |
|------|--------|
| `backend/app/routes/auth.py` | Rewrote `/auth/refresh` handler with insert-first atomic strategy |
| `backend/tests/test_auth.py` | Added 5 targeted tests for the new flow |

## Tests

| Test | What it verifies |
|------|-----------------|
| `test_refresh_returns_new_token_pair_on_valid_token` | Happy path: new pair issued, old JTI written to blacklist |
| `test_refresh_rejects_replayed_token_via_duplicate_key` | `DuplicateKeyError` on insert → 401, no tokens issued |
| `test_refresh_concurrent_requests_only_one_succeeds` | `asyncio.gather` with same token → exactly one 200, one 401 |
| `test_refresh_rejects_invalid_jwt` | Malformed token → 401 before any DB call |
| `test_refresh_rejects_access_token_used_as_refresh` | Wrong `type` claim → 401 |

All existing logout, blacklist, and lockout tests remain passing — no regressions to the normal refresh or logout flows.

## Security Impact

- **Before**: concurrent replay of a stolen refresh token could produce two valid token pairs from one rotation.
- **After**: exactly one caller wins the insert; the unique index guarantees no second caller receives tokens, regardless of timing or deployment topology.

### DIFF
--- a/backend/app/routes/auth.py
+++ b/backend/app/routes/auth.py
@@ -115,11 +115,20 @@ async def refresh(request: Request, body: RefreshRequest):
     """
     Exchange a valid refresh token for a new access + refresh token pair.
     Refresh token rotation: old refresh token is invalidated on use.
+
+    Atomicity guarantee: the old JTI is written to blacklisted_tokens via
+    insert_one BEFORE new tokens are issued.  The unique index on `jti`
+    (database.py) means a duplicate concurrent request raises DuplicateKeyError,
+    which is converted to 401 — eliminating the check-then-act race.
     """
-    username = await verify_refresh_token(body.refresh_token)
     ip_address = request.client.host if request.client else "unknown"
-    
-    if not username:
+
+    # ── Step 1: decode & basic structural validation (no DB round-trip yet) ──
+    try:
+        payload = jwt.decode(
+            body.refresh_token, settings.secret_key, algorithms=[ALGORITHM]
+        )
+    except JWTError:
         await _log_auth_event("unknown", ip_address, "refresh", False)
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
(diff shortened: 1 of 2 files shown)


Srejoye:
The tests which have failed are not from my PR. It was pre-existing.

Chetan0e (approved):
Thanks for the contribution.

This is a well-designed fix for a genuine security issue in refresh token rotation. The previous check-then-insert flow introduced a race condition that could allow concurrent reuse of the same refresh token.

Moving to an atomic insert-first blacklisting strategy is the correct approach here. Using the unique index on blacklisted_tokens.jti as the synchronization gate cleanly eliminates the race across concurrent requests and distributed deployments.

The added tests are strong and cover both functional and security-critical scenarios, especially replay and concurrent refresh handling.

I only recommend ensuring the unique index on blacklisted_tokens.jti is guaranteed to exist in all environments, since this implementation depends on it for correctness.

Overall, excellent fix. Approving this PR.

PR #205, opened 2026-06-29
fix: increment promoted_count/archived_count and fix silent-archival footgun in auto_promote_important_memories 

hariom888:
## Summary

Fixes #143.

`auto_promote_important_memories` initialised `promoted_count = 0` but never incremented it, so the guarded log line `if promoted_count > 0: logger.info(...)` was permanently dead code. Operators had no visibility into background lifecycle promotion activity.

A secondary bug was also present: any document that matched the MongoDB `$gte` filter but held a non-numeric `metadata.importance` value (e.g. `"high"` stored as a string) would return `0.0` from `.get()`, fail the Python-side re-check, and be **silently archived** — data loss with no
log trail.

---

## Changes

**`backend/app/db/memory_lifecycle.py`**

| Before | After |
|--------|-------|
| `promoted_count` initialised but never incremented | `promoted_count` and `archived_count` incremented in their respective branches using the `bool` return of `promote_to_long_term` / `archive_memory` |
| `if promoted_count > 0: logger.info(...)` — always skipped | `logger.info(...)` always emitted with both counts |
| Non-numeric `metadata.importance` silently archived | `float()` cast with `try/except`; emits `logger.warning` and skips the doc (`continue`) — no silent data loss |

---

## Verification against current main

Per the maintainer's request in the issue thread:

- **MongoDB query** — `metadata.importance: {$gte: 0.6}` with `lifecycle_stage: "short_term"` (unchanged).
- **Lifecycle state field** — `lifecycle_stage` (top-level), consistent with `promote_to_long_term`, `archive_memory`, and the existing schema tests in `test_lifecycle_schema.py`.
- **Python-side re-check necessity** — the DB filter already guarantees `>= 0.6` for well-typed documents; the re-check is kept only as a type-safety guard that now warns instead of silently archiving.
- **Counter placement** — `promoted_count += 1` inside the `if importance >= IMPORTANCE_THRESHOLD` branch; `archived_count += 1` inside the `else` branch.

---

## Tests (`backend/tests/test_auto_promote_counter.py`)

| Test | Asserts |
|------|---------|
| `test_promoted_count_incremented` | `promoted=2, archived=0` in log for 2 high-importance docs |
| `test_archived_count_incremented` | `archived=1, promoted=0` in log for 1 low-importance doc |
| `test_log_always_emitted_when_zero_promotions` | `"Auto-promotion complete"` present even with empty cursor |
[...]

PR #207, opened 2026-06-29
fix(lifecycle): canonicalize lifecycle field to top-level `lifecycle_stage`, fix stats aggregation and auto-promote counter

### LINKED ISSUE #114: bug: Lifecycle Field Name Divergence: `metadata.lifecycle` vs `lifecycle_stage` Causes Silent Lifecycle Failures
**Files:** `backend/app/services/memory_store.py`, `backend/app/db/memory_lifecycle.py`

### Description

Two subsystems that are supposed to jointly manage memory lifecycle state use different field names for the same concept, making them operate on entirely different document paths.

- `memory_store.py` reads/writes `metadata.lifecycle` (a nested field inside the `metadata` subdocument).
- `memory_lifecycle.py` (`MemoryLifecycleManager`) queries and updates `lifecycle_stage` (a top-level field).

```python
# memory_store.py:63 — stores lifecycle here
"lifecycle": metadata.get("lifecycle", "short_term")   # → doc.metadata.lifecycle

# memory_lifecycle.py:26 — queries a field that was never written
{"lifecycle_stage": "short_term"}  # → doc.lifecycle_stage (always missing)
```

### Consequence

- `promote_to_long_term`, `archive_memory`, `auto_promote_important_memories`, and `enforce_lifecycle_limits` all match zero documents because `lifecycle_stage` is never populated.
- `get_memories_by_stage` always returns empty lists, silently.
- The MongoDB index `(user_id, metadata.lifecycle)` created in `database.py:45` is used by `memory_store.py` searches but never by `MemoryLifecycleManager`, which would benefit from an index on `lifecycle_stage` that doesn't exist.

### Reproduction

1. Store any memory through the normal pipeline; confirm `metadata.lifecycle` = `"short_term"`.
2. Call `MemoryLifecycleManager.promote_to_long_term(user_id, memory_id)` — `modified_count` is 0 every time.
3. `get_memories_by_stage(user_id, "short_term")` returns `[]` despite records existing.

Srejoye:
## Summary

Fixes #114 — lifecycle field name divergence causing all `MemoryLifecycleManager` operations to silently match zero documents.

## Root Cause

The bug report described `memory_store.py` writing `metadata.lifecycle` and `memory_lifecycle.py` querying `lifecycle_stage`. Auditing the actual source reveals the real split:

| Location | Field written/queried | Status |
|---|---|---|
| `memory_store.store_memory` | `lifecycle_stage` (top-level) | ✅ Correct |
| `memory_store.store_memories_batch` | `lifecycle_stage` (top-level) | ✅ Correct |
| `memory_store.get_memory_stats` | `$metadata.lifecycle` (nested) | ❌ **Bug** — never written |
| `database.py` index | `(user_id, metadata.lifecycle)` | ❌ **Bug** — field doesn't exist |
| `memory_lifecycle.py` (all queries) | `lifecycle_stage` (top-level) | ✅ Already correct |
| `memory_lifecycle.auto_promote` | `promoted_count` never incremented | ❌ **Bug** — counter always 0 |
| `memory_lifecycle.auto_promote` | `else` archives docs that passed importance gate | ❌ **Bug** — wrong branch |

### `backend/app/services/memory_store.py`
- `get_memory_stats`: fixed aggregation `$group._id` from `"$metadata.lifecycle"` to `"$lifecycle_stage"`. The `by_lifecycle` bucket in stats was always `{short_term: 0, long_term: 0, archived: 0}` because the grouped field was never populated.

### `backend/app/services/database.py`
- Replaced index on `(user_id, metadata.lifecycle)` with `(user_id, lifecycle_stage)`. The old index was on a non-existent path, providing no query coverage.

### `backend/app/db/memory_lifecycle.py`
- `auto_promote_important_memories`: removed redundant inner importance re-check (documents already pass the `$gte` DB filter). Added `promoted_count += 1` after a successful `promote_to_long_term` call. Removed the `else: archive_memory()` branch that wrongly archived documents that had already passed the importance gate.

## Behavior Before / After

| Operation | Before | After |
|---|---|---|
| `get_memory_stats` `by_lifecycle` | Always `{short_term: 0, ...}` | Correct counts from `lifecycle_stage` field |
| `(user_id, lifecycle_stage)` index | Missing — queries unindexed | Index present, queries covered |
| `auto_promote` log message | Always "Auto-promoted 0 memories" | Accurate count of promoted documents |
| High-importance docs in `auto_promote` | Risk of wrong archival via `else` branch | Always promoted, never archived |

## Migration / Backfill

No data migration required. `memory_store.py` has written `lifecycle_stage` at the top level since the field was introduced. The bugs were purely in the read/query paths and index definition — no documents are malformed.

The stale `metadata.lifecycle` MongoDB index will be superseded by the new `lifecycle_stage` index on next application startup (when `setup_indexes` runs). A one-time `dropIndex` on `metadata.lifecycle` can be issued manually if index count is a concern, but it is not blocking.

## Tests

`backend/tests/test_lifecycle_field_divergence.py` — 11 new test cases covering:
- `store_memory` writes `lifecycle_stage` at top level, not `metadata.lifecycle`
- `get_memory_stats` aggregation pipeline groups on `$lifecycle_stage`
- `promote_to_long_term` filter and `$set` both use `lifecycle_stage`
- `get_memories_by_stage` queries `lifecycle_stage`
- `auto_promote` increments `promoted_count` accurately
- `auto_promote` does not archive high-importance documents
- `auto_promote` filter uses `metadata.importance`
- Full `short_term → long_term → archived` transition end-to-end
- `get_lifecycle_stats` counts by `lifecycle_stage`

### DIFF
--- a/backend/app/services/database.py
+++ b/backend/app/services/database.py
@@ -42,7 +42,7 @@ async def create_indexes():
         await db["memories"].create_index([("user_id", 1), ("created_at", -1)])
         await db["memories"].create_index([("user_id", 1), ("metadata.intent", 1)])
         await db["memories"].create_index([("user_id", 1), ("metadata.importance", -1)])
-        await db["memories"].create_index([("user_id", 1), ("metadata.lifecycle", 1)])
+        await db["memories"].create_index([("user_id", 1), ("lifecycle_stage", 1)])
         await db["memories"].create_index([("user_id", 1), ("metadata.speaker", 1)])
         await db["memories"].create_index("created_at", expireAfterSeconds=2592000)  # 30 days TTL for old documents if needed
 
--- a/backend/app/services/memory_store.py
+++ b/backend/app/services/memory_store.py
@@ -323,7 +323,7 @@ async def get_memory_stats(user_id: str) -> Dict[str, Any]:
     pipeline = [
         {"$match": {"user_id": user_id}},
         {"$group": {
-            "_id": "$metadata.lifecycle",
+            "_id": "$lifecycle_stage",
             "count": {"$sum": 1},
             "avg_importance": {"$avg": "$metadata.importance"},
         }}
--- a/backend/tests/test_lifecycle_field_divergence.py
+++ b/backend/tests/test_lifecycle_field_divergence.py
@@ -1,0 +1,332 @@
+"""
+Tests for lifecycle field name divergence.
+
+Verifies that:
+1. store_memory writes lifecycle_stage at the top level (not metadata.lifecycle).
+2. get_memory_stats aggregates on $lifecycle_stage (top-level), not $metadata.lifecycle.
+3. MemoryLifecycleManager.promote_to_long_term filters on top-level lifecycle_stage.
+4. MemoryLifecycleManager.get_memories_by_stage queries on top-level lifecycle_stage.
+5. auto_promote_important_memories increments promoted_count and does not archive
+   memories that passed the importance gate.
+6. Full short_term → long_term → archived lifecycle transition succeeds end-to-end.
+"""
+import pytest
+from unittest.mock import AsyncMock, MagicMock, patch, call
+
+
+# ── helpers ──────────────────────────────────────────────────────────────────
+
+def _make_collection(docs=None, modified_count=1, count_return=0):
+    """Build a mock Motor collection."""
+    col = MagicMock()
+    col.insert_one = AsyncMock(return_value=MagicMock())
+    col.insert_many = AsyncMock(return_value=MagicMock())
+    col.update_one = AsyncMock(return_value=MagicMock(modified_count=modified_count))
+    col.count_documents = AsyncMock(return_value=count_return)
+
+    async def _aiter(*a, **kw):
+        for d in (docs or []):
+            yield d
+
+    cursor = MagicMock()
+    cursor.__aiter__ = _aiter
+    cursor.sort = MagicMock(return_value=cursor)
+    cursor.limit = MagicMock(return_value=cursor)
+    col.find = MagicMock(return_value=cursor)
+
+    async def _aggregate(pipeline):
+        # simulate empty aggregate
+        return
+        yield  # make it an async generator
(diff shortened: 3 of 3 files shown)


Srejoye:
Hi @Chetan0e,

Chetan0e:
@Srejoye Thanks for the detailed investigation — the lifecycle field divergence fix is valid, and the fixes for `get_memory_stats` aggregation + DB index are definitely needed.

That said, I don’t think this should be merged as-is because it removes some safety/observability improvements introduced in PR #205 around `auto_promote_important_memories`.

Specifically, this PR removes:

* Python-side validation for malformed/non-numeric `metadata.importance`
* warning logs for bad records
* `archived_count`
* richer lifecycle logging

The Mongo filter on `metadata.importance >= threshold` is correct for clean data, but the additional Python-side guard from #205 protects against malformed records and improves debugging.

I’d recommend:

* keeping the lifecycle field fixes from this PR
* preserving the safety + observability improvements from #205
* rebasing after #205 (or updating this PR to include those changes)

So holding approval for now until those are reconciled.

Srejoye:
@Chetan0e Rebased onto current `main` and force-pushed. The old merge commit is gone, so `memory_lifecycle.py`, `auth.py`, `routes/auth.py`, `reminder_service.py`, and `digest.py` are now byte-identical to `main` again — none of the `#205`/`#206`/`#209`/`#214`/`#217` fixes are reverted anymore.

The PR is now scoped to just the two genuine field-name fixes plus the new test file:
- `memory_store.py`: `get_memory_stats` aggregation groups on `$lifecycle_stage`
- `database.py`: index uses `lifecycle_stage`
- `test_lifecycle_field_divergence.py`

Ready for re-review.

Chetan0e (approved):
Thanks for rebasing and narrowing the scope of the PR. The lifecycle field fixes and regression tests look good, and the previous concern about reverting improvements from #205 has been addressed. LGTM, approving.

PR #208, opened 2026-06-29
fix(websocket): validate token before accept() to eliminate unauthenticated connection window 

### LINKED ISSUE #115: security: WebSocket Endpoint Calls `websocket.accept()` Before Authentication, Leaking Connection State to Unauthenticated Clients
**File:** `backend/app/routes/websocket.py` (lines 53, 82)

### Description

The `/ws/updates` WebSocket handler accepts the connection unconditionally before any authentication takes place. The authentication frame is only processed *after* `accept()` succeeds. This means the server allocates a connection, performs I/O, and records the socket internally for an as-yet-unauthenticated client.

```python
# websocket.py:53 — connection accepted before identity is known
await websocket.accept()

# websocket.py:53-81 — auth frame received and validated AFTER accept
raw = await asyncio.wait_for(websocket.receive_text(), timeout=10.0)
...
user_id = await verify_access_token(token)
if not user_id:
    await websocket.close(code=4001, ...)

# websocket.py:82 — but now the unauthenticated socket also bypasses ConnectionManager.connect()
manager.active_connections[user_id] = websocket  # direct dict write, not via connect()
```

Additionally, `websocket_endpoint` registers the connection via direct dict assignment (`manager.active_connections[user_id] = websocket`) instead of `manager.connect()`, meaning the `ConnectionManager.connect()` method's `accept()` call and the endpoint's `accept()` call are duplicated across code paths, creating an inconsistency that could lead to double-accept errors if refactored.

### Impact

1. Unauthenticated WebSocket connections are briefly established and consume server file descriptors/memory — trivially exploitable for connection exhaustion with no rate limit applied at the transport layer.
2. Token timeout (10 s) means up to 10 s of open socket per unauthenticated client.
3. Any future path through `manager.connect()` would call `accept()` on an already-accepted socket, raising a Starlette error.

Srejoye:
## Summary

Fixes issue #115 — the `/ws/updates` WebSocket handler called `websocket.accept()` unconditionally before any authentication, allowing unauthenticated clients to hold open connections for up to 10 seconds each. Also eliminates the bypass of `ConnectionManager.connect()` that made the connection lifecycle inconsistent and double-accept-prone.

## Root Cause

| # | Problem | Location |
|---|---|---|
| 1 | `websocket.accept()` called before token validation | `websocket.py:53` |
| 2 | 10-second `receive_text()` window held unauthenticated socket open | `websocket.py:56` |
| 3 | Auth via first-frame message — unreachable before HTTP upgrade | design |
| 4 | Success path bypasses `manager.connect()` via direct dict write | `websocket.py:82` |
| 5 | `manager.connect()` would double-accept if ever called from endpoint | `websocket.py:17` |
| 6 | `broadcast()` iterated `active_connections` while potentially mutating it on failure | `websocket.py:33` |

### `backend/app/routes/websocket.py`

**Authentication flow:** Moved the JWT from a first-frame message field to the `token` query parameter (`wss://host/ws/updates?token=<jwt>`). Query parameters are part of the HTTP upgrade request and are available via `websocket.query_params` *before* `accept()` is called. If the token is missing or invalid, `websocket.close(code=4001)` is called pre-accept — Starlette converts this to an HTTP 403 during the upgrade negotiation. No file descriptor is allocated, no I/O buffers are consumed.

**Connection registration:** Removed the direct `manager.active_connections[user_id] = websocket` assignment. `manager.connect()` is now the single path that calls `accept()` and registers the socket, eliminating the inconsistency and the double-accept risk.

**Broadcast safety:** Fixed `broadcast()` to collect failed user IDs in a list and call `disconnect()` after the send loop completes, avoiding mutation of the dict being iterated.

## Behavior Before / After

| Scenario | Before | After |
|---|---|---|
| No token | `accept()` called → 10s timeout window → `close(4001)` | Pre-accept `close(4001)` → no FD allocated |
| Invalid token | `accept()` → receive frame → verify → `close(4001)` | Pre-accept `close(4001)` → no FD allocated |
| Auth service error | `accept()` → exception → `close(4001)` | Pre-accept `close(4001)` → no FD allocated |
| Valid token | `accept()` → direct dict write (bypasses `manager.connect()`) | Token verified → `manager.connect()` → `accept()` + register |
| Future `manager.connect()` call | Would double-accept → Starlette error | Not possible; single accept path |

## Security Note: Token in Query Parameters

Query parameters appear in web server access logs. This is the standard tradeoff for pre-accept WebSocket authentication. Recommended mitigations (outside scope of this PR):
- Configure log redaction for the `token` query parameter at the reverse proxy level
- Alternatively, implement a short-lived ticket pattern: client exchanges a REST-issued one-time token immediately before connecting, then uses the ticket in the query param

## Tests

`backend/tests/test_websocket_auth.py` — 16 test cases across 7 classes:
- Missing token → `close(4001)` without `accept()`, no stale connection
- Invalid token → `close(4001)` without `accept()`, no stale connection
- Auth service exception → `close(4001)` without `accept()`, no stale connection
- Valid token → `manager.connect()` called exactly once
- Valid token → `accept()` called exactly once (inside `manager.connect()`)
- Disconnect cleans up `active_connections`
- `ping` → `pong` response
- `send_personal_message` delivers, no-ops for unknown user, disconnects on send failure
- `broadcast` delivers to all users, removes dead connections without raising
- `ConnectionManager.connect()` is the sole `accept()` path (double-accept guard)

Closes #115

### DIFF
--- a/backend/app/routes/websocket.py
+++ b/backend/app/routes/websocket.py
@@ -14,6 +14,11 @@ class ConnectionManager:
         self.active_connections: dict[str, WebSocket] = {}
 
     async def connect(self, websocket: WebSocket, user_id: str):
+        """Accept the handshake and register the connection.
+
+        This is the single path that calls websocket.accept() — callers
+        must not call accept() themselves beforehand.
+        """
         await websocket.accept()
         self.active_connections[user_id] = websocket
         logger.info(f"WebSocket connected for user {user_id}")
@@ -33,11 +38,15 @@ class ConnectionManager:
 
     async def broadcast(self, message: dict):
         """Send message to all connected users."""
-        for user_id, connection in self.active_connections.items():
+        disconnected = []
+        for user_id, connection in list(self.active_connections.items()):
             try:
                 await connection.send_json(message)
             except Exception as e:
                 logger.error(f"Error broadcasting to {user_id}: {e}")
+                disconnected.append(user_id)
+        for user_id in disconnected:
+            self.disconnect(user_id)
 
 
 manager = ConnectionManager()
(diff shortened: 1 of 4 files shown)


Srejoye:
Hi @Chetan0e,

Chetan0e (changes requested):
Thanks for the contribution — this addresses a real security issue in the WebSocket auth flow.

Moving authentication before accept() eliminates the unauthenticated connection window and prevents clients from holding open unauthenticated sockets. The refactor to make ConnectionManager.connect() the single accept path also improves lifecycle consistency and removes double-accept risk. The broadcast iteration fix and added test coverage are both strong improvements.

I do have one blocking concern before merge: authentication is now performed using JWTs in query parameters (?token=<jwt>). While this enables pre-accept authentication, it introduces token exposure risk through access logs, reverse proxies, and monitoring infrastructure. This is especially important for production environments handling sensitive data.

Before merging, I’d like one of the following:

switch to a short-lived one-time websocket ticket approach, or
document and enforce query-param log redaction in deployment configuration

Once that concern is addressed, this PR looks solid.

Srejoye:
Hi @Chetan0e,

The CI failure here isn't caused by this PR's diff — it's failing at the dependency-install step, before any tests run.

`requirements.txt` pins `scipy>=1.18.0`, but `scipy` 1.18.0 requires Python ≥3.12, while the CI runner is on Python 3.11. Pip can't resolve any `scipy` version that satisfies both constraints, so `pip install -r requirements.txt` fails outright:

```
ERROR: No matching distribution found for scipy>=1.18.0
```

This is a repo-level CI/dependency issue, not something introduced by this change — it should reproduce on the base branch and on any other open PR that touches `backend/`.

Chetan0e (commented):
Thanks for the update. The CI issue appears unrelated to this PR, but my previous blocking review is still outstanding. The authentication flow now relies on JWTs in query parameters, and the requested mitigation (either a short-lived WebSocket ticket flow or deployment documentation for query-parameter log redaction) has not been addressed yet. Once that is implemented or documented, I'll be happy to review again.

PR #209, opened 2026-06-29
fix(auth): fail-closed token blacklist — deny revoked tokens when DB is unavailable

### LINKED ISSUE #142: Security: Token blacklist silently bypassed when database is unavailable during `verify_access_token`
### Summary
`verify_access_token` (and `verify_refresh_token`) in `backend/app/services/auth.py` (lines 54–70) only check the blacklist *if* `get_db()` returns a non-None value. When the database is temporarily unavailable — cold start, connection pool exhaustion, MongoDB restart — the blacklist check is skipped entirely and the token is accepted as valid.

### Reproduction
1. User logs out; their JWT `jti` is written to `blacklisted_tokens`.
2. MongoDB becomes unreachable (e.g., connection pool saturated, network partition).
3. The user replays the logged-out access token.
4. `get_db()` returns `None`; the blacklist branch is skipped.
5. `verify_access_token` returns the `username` — the revoked token is accepted.

### Root Cause
```python
# auth.py lines 63–66
db = get_db()
if db is not None:                              # ← silent bypass
    blacklisted = await db["blacklisted_tokens"].find_one({"jti": jti})
    if blacklisted:
        return None
# Falls through to: return payload.get("sub")  # ← accepts revoked token
```
The guard was added to tolerate cold-start ordering, but it creates a fail-open security path. The same pattern appears identically in `verify_refresh_token` (lines 85–88).

### Affected files
- `backend/app/services/auth.py` — `verify_access_token()` line 63–66, `verify_refresh_token()` lines 85–88

### Impact
An attacker who obtains a logged-out or rotated token can replay it during any period of database degradation. This undermines the entire token revocation model, particularly for logout and refresh rotation.

Srejoye:
## Summary

Closes #142

`verify_access_token` and `verify_refresh_token` both contained an `if db is not None` guard around the blacklist lookup. The intent was to handle cold-start ordering, but the same guard fires during *runtime* DB unavailability, silently skipping the revocation check and accepting any structurally valid JWT — including tokens invalidated by logout or refresh rotation.

## Root Cause

```python
# Before — fail-open
db = get_db()
if db is not None:                              # silent bypass on DB down
    blacklisted = await db["blacklisted_tokens"].find_one({"jti": jti})
    if blacklisted:
        return None
return payload.get("sub")                       # revoked token accepted
```

## Fix

Replace the fail-open guard with a fail-closed pattern in both functions:

1. **`get_db()` returns `None`** → log a `WARNING` (with `jti`) and return `None` immediately. Token denied.
2. **`find_one` raises** (pool exhaustion, network partition, etc.) → log an `ERROR` and return `None`. Token denied.
3. **DB available and JTI not blacklisted** → return username as before (no regression to happy path).

```python
# After — fail-closed
db = get_db()
if db is None:
    logger.warning("database unavailable — rejecting token (fail-closed); jti=%s", jti)
    return None
try:
    blacklisted = await db["blacklisted_tokens"].find_one({"jti": jti})
except Exception:
    logger.exception("blacklist lookup failed — rejecting token (fail-closed); jti=%s", jti)
    return None
if blacklisted:
    return None
```

No changes to startup lifecycle, `create_indexes`, connection pooling, or any other auth logic.

## Tests Added

| Test | Covers |
|---|---|
| `test_verify_access_token_rejects_when_db_unavailable` | `get_db()` → `None` on access token |
| `test_verify_refresh_token_rejects_when_db_unavailable` | `get_db()` → `None` on refresh token |
| `test_verify_access_token_rejects_when_blacklist_lookup_raises` | `find_one` exception on access token |
| `test_verify_refresh_token_rejects_when_blacklist_lookup_raises` | `find_one` exception on refresh token |
| `test_blacklisted_jti_rejected_even_after_db_recovers` | Outage then recovery — JTI still rejected |
| `test_valid_token_accepted_when_db_available_and_not_blacklisted` | Happy path not broken |

## Scope

Per the assignment, changes are confined to:
- `backend/app/services/auth.py` — `verify_access_token()`, `verify_refresh_token()`
- `backend/tests/test_auth.py` — new tests for the above

## Security Impact

Before this fix, any period of MongoDB degradation (cold start, pool exhaustion, restart, partition) allowed an attacker holding a logged-out or rotated JWT to replay it indefinitely. This fix ensures the revocation model holds regardless of DB availability.

### DIFF
--- a/backend/app/services/auth.py
+++ b/backend/app/services/auth.py
@@ -53,7 +53,13 @@ def create_refresh_token(username: str) -> str:
 
 # ── Token verification ────────────────────────────────────────────────────────
 async def verify_access_token(token: str) -> Optional[str]:
-    """Returns username if valid access token, else None."""
+    """Returns username if valid access token, else None.
+
+    Fails closed: if the database is unavailable (get_db() returns None)
+    or the blacklist lookup raises, the token is rejected rather than
+    silently accepted. This prevents revoked tokens from being replayed
+    during periods of DB degradation.
+    """
     try:
         payload = jwt.decode(token, settings.secret_key, algorithms=[ALGORITHM])
         if payload.get("type") != "access":
@@ -61,10 +67,22 @@ async def verify_access_token(token: str) -> Optional[str]:
         jti = payload.get("jti")
         if jti:
             db = get_db()
-            if db is not None:
+            if db is None:
+                logger.warning(
+                    "verify_access_token: database unavailable — rejecting token "
+                    "(fail-closed); jti=%s", jti
+                )
+                return None
+            try:
                 blacklisted = await db["blacklisted_tokens"].find_one({"jti": jti})
-                if blacklisted:
-                    return None
+            except Exception:
+                logger.exception(
+                    "verify_access_token: blacklist lookup failed — rejecting token "
+                    "(fail-closed); jti=%s", jti
+                )
+                return None
+            if blacklisted:
+                return None
         return payload.get("sub")
     except JWTError:
         return None
@@ -75,6 +93,10 @@ async def verify_refresh_token(token: str) -> Optional[str]:
 
     Also checks the blacklisted_tokens collection so that rotated (or
     explicitly revoked) refresh tokens cannot be reused.
+
+    Fails closed: if the database is unavailable (get_db() returns None)
+    or the blacklist lookup raises, the token is rejected rather than
+    silently accepted.
     """
     try:
         payload = jwt.decode(token, settings.secret_key, algorithms=[ALGORITHM])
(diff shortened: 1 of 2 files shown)


Srejoye:
Hi @Chetan0e,

Chetan0e (approved):
Thanks for the contribution — this fixes an important security issue in the authentication flow.

The previous blacklist verification logic failed open whenever get_db() returned None or the blacklist lookup failed, which meant revoked or rotated tokens could still be accepted during periods of database unavailability. That creates a serious security gap for logout and refresh-token rotation.

The new fail-closed approach is the correct behavior for authentication systems: if blacklist validation cannot be completed, the token should be rejected rather than accepted. I also appreciate that the fix consistently covers both verify_access_token() and verify_refresh_token() and includes solid test coverage for outage, exception, recovery, and happy-path scenarios.

There is a security-vs-availability tradeoff here (authenticated requests will fail during DB outages), but for revocation enforcement this is the safer default and aligns with expected auth behavior.

Looks good to merge.

PR #211, opened 2026-07-01
fix(tests): fix 4 bugs in test_system.py - nodes field, signup 201, None crash, false success message

### LINKED ISSUE #191: fix(tests): test_status() references non-existent 'nodes' field in status response
**Describe the bug**
In `test_status()`:

```python
print_result(True, f"System is {data.get('status')} with {data.get('nodes')} nodes")
```

The `/status` endpoint returns:
```json
{"status": "ok", "version": "3.0.0", "scheduler": "running"}
```

There is no `nodes` field. `data.get('nodes')` silently returns `None`, so the output reads:
 PASS: System is ok with None nodes

**Expected behavior**
Should reference an actual response field:
```python
print_result(True, f"System is {data.get('status')}, version {data.get('version')}, scheduler {data.get('scheduler')}")
```

Tech4Aditya:
I have found more issues linked with the open issue, to avoid any conflicts i have applied the following changes,

Fixes #191

### Fix 1 — `test_status()`: replaced non-existent `nodes` field (Issue #191)
`data.get('nodes')` always returned `None` since the `/status` endpoint
returns `status`, `version`, and `scheduler` — not `nodes`.

**Before:**
```python
print_result(True, f"System is {data.get('status')} with {data.get('nodes')} nodes")
```
**After:**
```python
print_result(True, f"System is {data.get('status')}, version {data.get('version')}, scheduler {data.get('scheduler')}")
```

---

### Fix 2 — `test_signup()`: accept HTTP 201 from `/auth/signup`
The API returns `201 Created` on successful signup (as documented in README),
but the test only accepted `200`. This caused the entire test suite to abort
on a fully working backend since `username` returned `None` and the early-exit
guard triggered.

**Before:**
```python
if response.status_code == 200:
```
**After:**
```python
if response.status_code in (200, 201):
```

---

### Fix 3 — `test_query()` and `test_summary()`: guard against `None` crash on `[:50]` slice
When a new user has zero memories, the backend returns `{"answer": null}`.
`data.get('answer')[:50]` raises `TypeError: 'NoneType' object is not subscriptable`,
crashing the test instead of reporting a clean result. Same bug existed in
`test_summary()` with the `summary` field — fixed both.

**Before:**
```python
print_result(True, f"Query returned: {data.get('answer')[:50]}...")
```
**After:**
```python
answer = data.get('answer') or ""
print_result(True, f"Query returned: {answer[:50]}...")
```

---

### Fix 4 — `main()`: track test results, don't unconditionally print success
The final `"✅ All basic endpoints are functional!"` message printed regardless
of whether any tests failed. Return values from all test functions were being
discarded. Now collected into a `results` list with an accurate pass/fail summary.

**Before:**
```python
test_query(token)
# ...
print("\n✅ All basic endpoints are functional!")
```
**After:**
```python
results = []
results.append(test_query(token))
# ...
failed = results.count(False)
if failed == 0:
    print("\n✅ All basic endpoints are functional!")
else:
    print(f"\n⚠️  {failed} test(s) failed. Check output above.")
```

---

## Files Changed
- `backend/test_system.py`

### DIFF
--- a/backend/test_system.py
+++ b/backend/test_system.py
@@ -27,7 +27,8 @@ def test_status():
         response = requests.get(f"{BASE_URL}/status")
         if response.status_code == 200:
             data = response.json()
-            print_result(True, f"System is {data.get('status')} with {data.get('nodes')} nodes")
+            # FIX 1: replaced non-existent 'nodes' field with actual response fields
+            print_result(True, f"System is {data.get('status')}, version {data.get('version')}, scheduler {data.get('scheduler')}")
             return True
         else:
             print_result(False, f"Status code: {response.status_code}")
@@ -45,7 +46,8 @@ def test_signup():
             f"{BASE_URL}/auth/signup",
             json={"username": username, "password": "testpass123"}
         )
-        if response.status_code == 200:
+        # FIX 2: /auth/signup returns 201 Created, not 200
+        if response.status_code in (200, 201):
             data = response.json()
             print_result(True, f"User created: {username}")
             return username
@@ -86,7 +88,9 @@ def test_query(token):
         )
         if response.status_code == 200:
             data = response.json()
-            print_result(True, f"Query returned: {data.get('answer')[:50]}...")
+            # FIX 3: guard against None answer crashing with TypeError on [:50]
+            answer = data.get('answer') or ""
+            print_result(True, f"Query returned: {answer[:50]}...")
             return True
         else:
             print_result(False, f"Status code: {response.status_code}, Error: {response.text}")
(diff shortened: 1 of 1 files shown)


Tech4Aditya:
@Chetan0e Kindly review and merge the PR

Chetan0e (approved):
Thanks for the contribution — this PR fixes multiple legitimate issues in test_system.py and improves the reliability of the test suite without affecting production code.

The fixes for the invalid nodes field, signup status code mismatch (201 vs 200), and NoneType crashes in query/summary tests all address real bugs that could cause false negatives or unexpected test crashes.

The final summary fix is also valuable since the suite previously printed a success message even when tests failed, which made results misleading.

One minor thing to verify before merge: please ensure all test functions consistently return explicit boolean values (True / False) so the aggregated result summary remains accurate.

Overall, good improvement to test stability. Looks good to merge.

PR #212, opened 2026-07-01
fix: query created_at not timestamp in digest active-user lookup (#202)

### LINKED ISSUE #202: Bug: weekly digest scheduler silently processes zero users - `_get_active_user_ids` filters the `timestamp` string field with a datetime, which never matches in MongoDB
## Describe the bug

`run_weekly_digests()` (the scheduled job that generates weekly digests for active users) silently processes zero users on every run, because `_get_active_user_ids()` queries the wrong field with a type that can never match.

In `app/services/digest.py`:

```python
async def _get_active_user_ids(window_hours: int) -> List[str]:
    db = get_db()
    if db is None:
        return []
    cutoff = datetime.utcnow() - timedelta(hours=window_hours)
    user_ids = await db["memories"].distinct(
        "user_id", {"timestamp": {"$gte": cutoff}}    # ← string field, datetime value
    )
    return [str(uid) for uid in user_ids if uid]
```

The problem is a field-type mismatch. When memories are stored (`app/services/memory_store.py`), the document has two different time fields:

```python
mongo_doc = {
    ...
    "timestamp": timestamp.isoformat(),   # stored as a STRING, e.g. "2026-06-29T10:00:00"
    "created_at": timestamp,              # stored as a datetime (BSON Date)
    "updated_at": timestamp,
}
```

`_get_active_user_ids` filters `{"timestamp": {"$gte": cutoff}}` where `cutoff` is a `datetime`. In MongoDB, `$gte` only compares values of the same BSON type — a `Date` predicate does not match `String`-typed fields. Since `timestamp` is stored as an ISO string, the query matches **no documents**, so `distinct("user_id", ...)` always returns an empty list.

The result: `run_weekly_digests()` iterates over an empty list and creates zero digests every week, even when users have recent activity. The feature appears wired up (scheduler job registered, endpoints exist) but never actually produces a scheduled digest.

## Steps to Reproduce

1. Have at least one user with memories created in the last 7 days.
2. Trigger the weekly digest job: `await run_weekly_digests()` (or wait for the scheduled run).
3. Observe the log line `Weekly digest job complete: 0 digests generated` despite active users existing.
4. `GET /digest/latest` continues to return 404 for those users.

Quick confirmation that the query returns nothing:

```python
from datetime import datetime, timedelta
cutoff = datetime.utcnow() - timedelta(hours=168)
# 'timestamp' is stored as "2026-06-29T10:00:00" (string); cutoff is a datetime
await db["memories"].distinct("user_id", {"timestamp": {"$gte": cutoff}})  # -> []
```

## Expected Behaviour

`_get_active_user_ids` should return every user with memory activity in the window, so the digest job generates one digest per active user.

## Actual Behaviour

It returns an empty list every time, so no scheduled digests are ever generated. The manual `POST /digest/generate` endpoint still works (it takes the user_id directly and doesn't go through this query), which is why the bug is easy to miss — only the scheduled path is broken.

## Suggested Fix

Filter on `created_at`, the BSON `Date` field that every other query in the codebase already uses (e.g. `get_memory_stats` and `all_memories_filtered` both filter `created_at` with datetime values):

```python
user_ids = await db["memories"].distinct(
    "user_id", {"created_at": {"$gte": cutoff}}
)
```

This is a one-line change. The docstring on `_get_active_user_ids` also says it "uses the same `timestamp` field that the summarizer's timeline query filters on" — but `get_recent_timeline` actually filters on `created_at` (see `app/services/timeline.py`), so the docstring is describing behaviour the code doesn't implement. Worth updating the docstring to match.

I can put up a PR with the fix plus a test that asserts an active user is returned by `_get_active_user_ids` within the window if that looks right.

Yuthika10:
## Summary

Fixes #202. The weekly digest scheduler silently generated zero digests every run because `_get_active_user_ids` queried the wrong field type in MongoDB.

## The bug

`_get_active_user_ids` in `app/services/digest.py` filtered:

```python
{"timestamp": {"$gte": cutoff}}   # cutoff is a datetime
```

But memories store two different time fields (`app/services/memory_store.py`):

- `timestamp` as an ISO **string** (`timestamp.isoformat()`)
- `created_at` as a BSON **Date** (datetime)

MongoDB's `$gte` only compares values of the same BSON type, so a datetime predicate never matches the string `timestamp` field. The query returned an empty list on every run, so `run_weekly_digests()` iterated over nothing and generated zero digests, even when users had recent activity.

The manual `POST /digest/generate` path was unaffected, since it takes `user_id` directly and doesn't go through this query, which is why the bug wasn't obvious.

## The fix

Changed the filter to `created_at`, the Date field that the other memory queries in the codebase already use (`get_memory_stats`, `all_memories_filtered`):

```python
{"created_at": {"$gte": cutoff}}
```

Also corrected the docstring, which claimed the function matched the summarizer's timeline field but described the wrong one (`get_recent_timeline` reads `created_at`).

## Tests

Added a test asserting `_get_active_user_ids` queries `created_at` and explicitly not `timestamp`, so this mismatch can't silently return. Existing digest tests still pass (11 total).

### DIFF
--- a/backend/app/services/digest.py
+++ b/backend/app/services/digest.py
@@ -14,15 +14,17 @@ DEFAULT_WINDOW_HOURS = 168
 async def _get_active_user_ids(window_hours: int) -> List[str]:
     """Return user_ids with memory activity in the window.
 
-    Uses the same `timestamp` field that the summarizer's timeline query
-    filters on, so we only digest users the summary would actually cover.
+    Filters on `created_at`, the BSON Date field used by the other memory
+    queries in the codebase (get_memory_stats, all_memories_filtered). The
+    `timestamp` field is stored as an ISO string, so a datetime `$gte`
+    predicate against it never matches in MongoDB.
     """
     db = get_db()
     if db is None:
         return []
     cutoff = datetime.utcnow() - timedelta(hours=window_hours)
     user_ids = await db["memories"].distinct(
-        "user_id", {"timestamp": {"$gte": cutoff}}
+        "user_id", {"created_at": {"$gte": cutoff}}
     )
     return [str(uid) for uid in user_ids if uid]
 
--- a/backend/tests/test_digest.py
+++ b/backend/tests/test_digest.py
@@ -125,6 +125,25 @@ class TestDigestService:
         assert generated_for == ["user_1", "user_2"]
         memories.distinct.assert_called_once()
 
+    async def test_get_active_user_ids_queries_created_at_not_timestamp(self, monkeypatch):
+        """The active-user query must filter on created_at (BSON Date), not the
+        timestamp string field, or it silently matches nothing in MongoDB."""
+        memories = MagicMock()
+        memories.distinct = AsyncMock(return_value=["user_1"])
+        db = _make_db({"memories": memories})
+        monkeypatch.setattr("app.services.digest.get_db", lambda: db)
+
+        from app.services.digest import _get_active_user_ids
+
+        result = await _get_active_user_ids(168)
+
+        assert result == ["user_1"]
+        memories.distinct.assert_called_once()
+        field, query = memories.distinct.call_args.args
+        assert field == "user_id"
+        assert "created_at" in query
+        assert "timestamp" not in query
+
     async def test_run_weekly_digests_skips_inactive_users(self, monkeypatch):
         """With no active users, no digests are generated."""
         memories = MagicMock()


Yuthika10:
Hi @Chetan0e fixed #202 using this PR. Ready for review

Chetan0e (approved):
Thanks for the contribution — this fixes a legitimate production issue in weekly digest generation.

The root cause is the mismatch between querying timestamp (stored as ISO string) using a Python datetime, which causes MongoDB to return no matches due to BSON type mismatch. As a result, _get_active_user_ids() silently returned an empty list and weekly digests were never generated for active users.

Switching the query to created_at is the correct fix since it is stored as a BSON Date and aligns with the rest of the memory query logic.

The regression test is also good and ensures future changes won’t accidentally revert back to querying timestamp.

One thing worth verifying before merge: ensure older memory documents also contain created_at, otherwise some legacy data may still be skipped.

Overall, solid fix. Looks good to merge.

PR #214, opened 2026-07-02
fix: route auth/reminder DB access through shared get_db() pool

### LINKED ISSUE #190: [BUG] `auth.py` and `reminder_service.py` each open their own MongoDB client at import time, bypassing the shared connection pool
## Description

Files: `app/services/auth.py` (`_mongo = AsyncIOMotorClient(...)`, `_users_col`, `_login_attempts_col`) and `app/services/reminder_service.py` (`_memories_col`, `_alerts_col`) each create a brand-new `AsyncIOMotorClient` at module load, completely independent of `app/services/database.py`'s pooled `db_instance` and its configured `maxPoolSize`/timeouts. These connections are never closed by `close_mongo_connection()`, and a Mongo outage at import time crashes startup the same way #112 describes. It also means `get_db()`-based mocking elsewhere in the app doesn't apply to auth/reminders - three different, inconsistent DB-access patterns in one codebase.

**Fix direction:** route auth/reminder services through the shared `get_db()` instead of their own clients.

**Covers:** root cause behind several auth/reminder test failures.

**Labels:** `type:bug`, `gssoc`

payalrvs3:
Closes #190

---

### Problem

`app/services/auth.py` and `app/services/reminder_service.py` each called `AsyncIOMotorClient(settings.mongo_uri)` at **module import time**, creating connections that were:

- **Never closed** by `close_mongo_connection()` on shutdown
- **Outside the shared pool** - `maxPoolSize`, timeouts configured in `database.py` had no effect on these clients
- **Crash-at-startup** if Mongo was unreachable at import time 
- **Untestable via `get_db()` mocks** - three different DB-access patterns in one codebase

### Fix

Removed all module-level `AsyncIOMotorClient` instantiation from both files. Every collection access now goes through `get_db()["collection_name"]` inside the function body, consistent with `digest.py` and the rest of the app.

**`auth.py`** - removed `_mongo`, `_users_col`, `_login_attempts_col`; `get_db()` called inside `create_user`, `authenticate_user`,
`get_lockout_seconds_remaining`, `register_failed_login`, `reset_login_attempts`.

**`reminder_service.py`** - removed `_mongo`, `_db`, `_memories_col`, `_alerts_col`; `get_db()` called inside `check_and_fire_reminders`, `get_upcoming_reminders`, `acknowledge_reminder`.

### Test changes

Three test files patched the now-removed module-level variables directly
(`_users_col`, `_login_attempts_col`, `_memories_col`, `_alerts_col`).
These were updated to patch `get_db` instead - necessary consequence of the fix, not refactoring.

| File | Change |
|---|---|
| `tests/test_account_lockout.py` | patch `get_db` instead of `_login_attempts_col` |
| `tests/test_system.py` | patch `get_db` instead of `_users_col` |
| `tests/test_reminders.py` | patch `get_db` instead of `_memories_col`/`_alerts_col` |

### Pre-existing failure (not caused by this PR)

`test_login_route_returns_429_after_threshold` fails because `exception_handlers.py`'s custom `http_exception_handler` builds its own `JSONResponse` without forwarding `exc.headers`, silently dropping the `Retry-After` header on every `HTTPException` in the app. This exists on `main` independently of this PR and is unrelated to connection pooling.

### Notes

- `memory_store.py` has the same standalone-client pattern - intentionally left out of scope per maintainer guidance, tracked separately
- No behaviour changes - pure architectural fix routing DB access through the existing shared lifecycle managed by `database.py`

### DIFF
--- a/backend/app/services/auth.py
+++ b/backend/app/services/auth.py
@@ -7,18 +7,12 @@ import bcrypt
 from fastapi import Depends, HTTPException, status
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 from jose import JWTError, jwt
-from motor.motor_asyncio import AsyncIOMotorClient
 
 from app.config import settings
 from app.services.database import get_db
 
 
 logger = logging.getLogger(__name__)
-
-_mongo = AsyncIOMotorClient(settings.mongo_uri)
-_db = _mongo[settings.database_name]
-_users_col = _db["users"]
-_login_attempts_col = _db["login_attempts"]
 
 ACCESS_TOKEN_EXPIRE_MINUTES = 10080  # 7 days in minutes
 REFRESH_TOKEN_EXPIRE_DAYS = 30
@@ -104,10 +98,11 @@ def decode_access_token(token: str) -> Optional[Dict]:
 # ── User operations ───────────────────────────────────────────────────────────
 async def create_user(username: str, password: str) -> bool:
     username = username.lower().strip()
-    existing = await _users_col.find_one({"username": username})
+    db = get_db()
+    existing = await db["users"].find_one({"username": username})
     if existing:
         return False
-    await _users_col.insert_one({
+    await db["users"].insert_one({
         "username": username,
         "password_hash": hash_password(password),
         "created_at": datetime.utcnow(),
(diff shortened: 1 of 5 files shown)


Chetan0e (approved):
Thanks for the contribution — this is a solid architectural fix and improves consistency across the codebase.

Removing module-level AsyncIOMotorClient instances from auth and reminder services ensures all MongoDB access now flows through the shared get_db() connection lifecycle. This fixes unmanaged pools, startup-time connection risks, and improves testability by standardizing mocking through get_db().

I also appreciate that the PR keeps scope tight with no behavioral changes to auth or reminder logic. The test updates are aligned with the refactor and make the suite cleaner.

The noted Retry-After header issue in exception_handlers.py appears unrelated and can be handled separately.

Looks good to merge.

PR #217, opened 2026-07-04
fix: normalize datetimes to UTC-aware in reminder scan to prevent tz comparison crash (#215)

### LINKED ISSUE #215: check_and_fire_reminders crashes with "can't compare offset-naive and offset-aware datetimes" when a memory's date string contains a Z/UTC offset, aborting the entire reminder scan
Hi again! I found a datetime comparison bug in the reminder scheduler while reading `backend/app/services/reminder_service.py`.

## The bug

In `check_and_fire_reminders`, `now` is a naive datetime:

```python
now = datetime.utcnow()                      # naive (no tzinfo)
window_end = now + timedelta(hours=LOOKAHEAD_HOURS)   # naive
```

But when a stored date string ends in `Z` (or carries an explicit offset), `datetime.fromisoformat` produces a timezone-aware datetime:

```python
try:
    if isinstance(parsed_str, str):
        parsed_date = datetime.fromisoformat(parsed_str.replace("Z", "+00:00"))  # AWARE for "...Z"
    else:
        continue
except (ValueError, TypeError):
    # dateparser fallback → returns a NAIVE datetime
    ...
```

The window check then compares the naive `now`/`window_end` against the aware `parsed_date`, and this line is outside any try/except:

```python
if not (now <= parsed_date <= window_end):   # TypeError when parsed_date is aware
    continue
```

In Python 3 that raises:
`TypeError: can't compare offset-naive and offset-aware datetimes`

Because the comparison isn't guarded, the exception propagates out of the `async for` loop and aborts the whole `check_and_fire_reminders` run. The scheduler calls this every 15 minutes, so as soon as the memories collection contains one date stored with a `Z`/offset, every reminder scan dies on that memory and no reminders fire for anyone.

The mixed-format situation is realistic here: the ISO-with-`Z` path yields aware datetimes, while the `dateparser` fallback (used for phrases like "next Monday") yields naive ones, so the same collection produces both kinds and the comparison is inconsistent.

## Reproduce

```python
from datetime import datetime, timedelta

now = datetime.utcnow()                       # naive
window_end = now + timedelta(hours=24)        # naive
parsed_date = datetime.fromisoformat("2026-07-05T14:00:00Z".replace("Z", "+00:00"))  # aware

now <= parsed_date <= window_end
# TypeError: can't compare offset-naive and offset-aware datetimes
```

## Suggested fix

Normalize everything to a single convention. Simplest is to make `now` timezone-aware and coerce any naive `parsed_date` to UTC before comparing:

```python
from datetime import datetime, timezone

now = datetime.now(timezone.utc)                        # aware UTC
window_end = now + timedelta(hours=LOOKAHEAD_HOURS)

# after parsing parsed_date (either branch):
if parsed_date.tzinfo is None:
    parsed_date = parsed_date.replace(tzinfo=timezone.utc)   # treat naive as UTC
```

This makes both sides aware so the window comparison is always valid, regardless of whether the date came from the ISO-with-`Z` path or the dateparser fallback. (`get_upcoming_reminders` in the same file also uses `datetime.utcnow()` for its `alerted_at` window; worth making it consistent, though it doesn't compare against aware values so it isn't currently crashing.)

Can put up a PR with the fix and a test that feeds a `...Z` date through `check_and_fire_reminders` and asserts it doesn't raise if that looks right.

Yuthika10:
## Summary

Fixes #215. `check_and_fire_reminders` crashed with `TypeError: can't compare offset-naive and offset-aware datetimes` whenever a memory's date string carried a `Z`/UTC offset, aborting the entire reminder scan.

## The bug

In `backend/app/services/reminder_service.py`, `now` was naive (`datetime.utcnow()`), but `datetime.fromisoformat(parsed_str.replace("Z", "+00:00"))` produces a timezone-aware datetime for any `...Z` date. The window check `now <= parsed_date <= window_end` compared naive against aware and raised. Since that comparison wasn't guarded, the exception propagated out of the `async for` loop and killed the whole scan. The scheduler runs every 15 minutes, so a single `Z`/offset date in the memories collection meant no reminders fired for anyone.

The mixed-format situation is real: the ISO-with-`Z` path yields aware datetimes while the `dateparser` fallback (for phrases like "next Monday") yields naive ones, so the same collection produces both.

## The fix

Normalized everything to a single convention (UTC-aware), scoped to `reminder_service.py`:

- `now` is now `datetime.now(timezone.utc)` (aware), so `window_end` is aware too.
- After parsing (either branch), a naive `parsed_date` is coerced to UTC with `parsed_date.replace(tzinfo=timezone.utc)` before the comparison.

This makes both sides of the window check (and the `parsed_date - now` arithmetic) always aware, regardless of which parse path produced the date.

I also changed `now` in `get_upcoming_reminders` to `datetime.now(timezone.utc)` for consistency. Since `check_and_fire_reminders` now stores `alerted_at` as an aware datetime, keeping the `get_upcoming_reminders` query bound naive would create the same naive-vs-aware mismatch on the `alerted_at` `$gte` filter, so making it aware keeps the whole file on one convention. Happy to revert that one line if you'd prefer to keep the change even more minimal.

## Tests

Added two regression tests to `tests/test_reminders.py`, matching the existing mock style:

- `test_z_suffixed_date_does_not_crash_comparison` — feeds a `...Z` (aware) date through `check_and_fire_reminders` and asserts it doesn't raise and the alert fires. This fails on the current code with the reported `TypeError`.
- `test_naive_date_still_processed_without_error` — feeds a naive date (the dateparser-fallback shape) and asserts it's processed without error.

Both pass with the fix. (The three existing endpoint tests in this file fail locally with 401 because they need a live database for token verification, which fails closed without one — that's pre-existing and unrelated to this change; they pass in CI.)

### DIFF
--- a/backend/app/services/reminder_service.py
+++ b/backend/app/services/reminder_service.py
@@ -1,5 +1,5 @@
 import logging
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, List, Optional
 
 from app.services.database import get_db
@@ -24,7 +24,7 @@ async def check_and_fire_reminders() -> int:
     memories_col = db["memories"]
     alerts_col = db["alerts"]
 
-    now = datetime.utcnow()
+    now = datetime.now(timezone.utc)
     window_end = now + timedelta(hours=LOOKAHEAD_HOURS)
 
     cursor = memories_col.find({
@@ -66,6 +66,9 @@ async def check_and_fire_reminders() -> int:
                         continue
                 except Exception:
                     continue
+
+            if parsed_date.tzinfo is None:
+                parsed_date = parsed_date.replace(tzinfo=timezone.utc)
 
             if not (now <= parsed_date <= window_end):
                 continue
@@ -117,7 +120,7 @@ async def get_upcoming_reminders(
     db = get_db()
     alerts_col = db["alerts"]
 
-    now = datetime.utcnow()
+    now = datetime.now(timezone.utc)
 
     query: Dict[str, Any] = {
         "user_id": user_id,
--- a/backend/tests/test_reminders.py
+++ b/backend/tests/test_reminders.py
@@ -44,6 +44,85 @@ class TestReminders:
         assert count >= 1
         mock_alerts_col.insert_one.assert_called()
 
+    async def test_z_suffixed_date_does_not_crash_comparison(self, monkeypatch):
+        """A memory date ending in 'Z' (timezone-aware) must not raise
+        'can't compare offset-naive and offset-aware datetimes'."""
+        from datetime import datetime, timedelta, timezone
+
+        # Aware future date, serialized with a trailing Z (the crashing case)
+        future_z = (
+            datetime.now(timezone.utc) + timedelta(hours=1)
+        ).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+        mock_memories_col = MagicMock()
+        mock_memories_col.find = MagicMock(return_value=_async_cursor([
+            {
+                "_id": "mem_z",
+                "user_id": "test_user",
+                "text": "Meeting soon",
+                "metadata": {
+                    "intent": "meeting",
+                    "entities": {"dates": [future_z]},
+                },
+            }
+        ]))
+
+        mock_alerts_col = MagicMock()
+        mock_alerts_col.find_one = AsyncMock(return_value=None)
+        mock_alerts_col.insert_one = AsyncMock()
+
+        mock_db = MagicMock()
+        mock_db.__getitem__ = MagicMock(side_effect=lambda name: {
+            "memories": mock_memories_col,
+            "alerts": mock_alerts_col,
+        }.get(name, MagicMock()))
+        monkeypatch.setattr("app.services.reminder_service.get_db", lambda: mock_db)
+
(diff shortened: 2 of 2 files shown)


Yuthika10:
Hi @Chetan0e!
Heads up on the failing CI: it doesn't look related to this change.

The install step fails on `scipy>=1.18.0` (`backend/requirements.txt` line 20), because scipy 1.18.0+ requires Python >=3.12, but the workflow runs Python 3.11 (the `Set up Python 3.11` step). So dependency install fails before any tests run, which would affect any PR currently, not just this one.

This PR only touches `backend/app/services/reminder_service.py` and `backend/tests/test_reminders.py` and doesn't change any dependencies.

Locally on Python 3.11, the two new regression tests (`test_z_suffixed_date_does_not_crash_comparison`, `test_naive_date_still_processed_without_error`) and the existing reminder-alert test pass. The three endpoint tests in the same file fail only because there's no database available for token verification, so auth fails closed with a 401 - that's environmental and unrelated to this change.

Happy to open a separate issue for the scipy / Python-version pin if that would help.

Chetan0e (approved):
Looks good to me — this fixes the timezone comparison issue in reminder scanning and adds solid regression coverage for both timezone-aware and naive datetime cases. The approach is clean, scoped, and addresses the root cause well. Nice work
Approved.

PR #231, opened 2026-07-11
Fix WebSocket disconnect to target the specific closing connection

### LINKED ISSUE #229: [Bug] WebSocket disconnect has no connection-identity check
## Description

Connections are tracked one-per-user with no way to tell two connections for the same user apart. If a user briefly has two overlapping connections (e.g. a quick reconnect), the older one disconnecting can silently knock the newer, still-active connection out of routing, with no error on either side.

**Location:** `backend/app/routes/websocket.py`

**Labels:** gssoc, type:bug

payalrvs3:
## Summary
Fixes #229. Connections were tracked as `Dict[user_id, WebSocket]`, so a user could only ever have one connection on record. The endpoint also bypassed `connect()`, writing to `active_connections` inline, and `disconnect(user_id)` deleted by user_id alone with no check on which socket was actually closing. During a quick reconnect, the older connection closing would silently knock the newer, still-active one out of routing.

`active_connections` is now `Dict[user_id, Set[WebSocket]]`. `disconnect()` takes the specific websocket and removes only that one, leaving any others intact. The endpoint routes registration through `manager.connect()` instead of mutating the dict inline, and `send_personal_message`/`broadcast` now iterate every connection a user holds instead of assuming one.

No changes to the wire-level API - the auth handshake and ping/pong are untouched. `ConnectionManager` isn't imported anywhere outside this file, so nothing else depends on the old `disconnect(user_id)` signature.

## Test plan
- [x] `pytest tests/test_websocket_connection_identity.py -v` - 8 new tests covering multi-connection tracking, targeted disconnect, and message fan-out to all of a user's sockets. All passing.
- [x] Confirmed against the unmodified code that the core regression test fails without this fix (an older socket's disconnect wipes out a newer, still-active one), confirming it's a genuine regression test.

## Screenshots (if UI)
N/A - backend-only fix, no UI changes.

### DIFF
--- a/backend/app/routes/websocket.py
+++ b/backend/app/routes/websocket.py
@@ -8,36 +8,46 @@ router = APIRouter()
 
 
 class ConnectionManager:
-    """Manages WebSocket connections for real-time updates."""
+    """Manages WebSocket connections for real-time updates.
+
+    Tracks a set of connections per user_id so a user can briefly hold more
+    than one active connection, e.g. during a reconnect.
+    """
 
     def __init__(self):
-        self.active_connections: dict[str, WebSocket] = {}
+        self.active_connections: dict[str, set[WebSocket]] = {}
 
     async def connect(self, websocket: WebSocket, user_id: str):
-        await websocket.accept()
-        self.active_connections[user_id] = websocket
+        """Register an already-accepted socket under user_id (caller must accept() first)."""
+        self.active_connections.setdefault(user_id, set()).add(websocket)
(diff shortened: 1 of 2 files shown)


Chetan0e:
@payalrvs3 Thanks for the contribution. This addresses the reconnect race condition cleanly by tracking connections per user and removing only the specific socket that disconnects. The centralized connection management and added regression tests are solid improvements. I reviewed the changes and they look good. LGTM, approving.

Chetan0e (approved):

PR #245, opened 2026-07-17
fix: execute persistent queue tasks via a consumer loop

### LINKED ISSUE #228: [Bug] Persistent task queue never executes enqueued tasks
## Description

Tasks submitted through the background queue (used by /record/session and /lifecycle/compress) are written to the database and reported as successful, but nothing in the codebase ever consumes and runs them. Session-based recordings never get transcribed, and the memory-lifecycle promotion/archival logic has no other entry point, so it never runs either - automatically or manually.

**Location:** `backend/app/workers/background_worker.py`, `backend/app/workers/task_queue.py`

**Labels:** gssoc, type:bug

payalrvs3:
Closes #228

### Root cause
`enqueue_task()` stored only a string preview of args/kwargs instead of reconstructable data, and nothing in the codebase ever called `task_queue.dequeue()`. Tasks were written to Mongo and reported as successful, but never picked up or run - so recordings never got transcribed and memory-lifecycle compression never ran.

### Fix
- `enqueue_task()` now takes `task_type` + a serializable `payload` dict instead of an unstorable closure, so a task can be reconstructed after a restart.
- Added a `TASK_HANDLERS` dispatch table mapping `RECORDING`/`COMPRESSION` to their handler functions.
- Added `_consumer_loop()`, which polls `dequeue()` and reports outcomes through the existing `update_status`/`mark_for_retry` API - retry (exponential backoff) and dead-letter logic in `task_queue.py` is
  unchanged.
- `start_worker()`/`stop_worker()` are wired into the FastAPI lifespan, starting after Mongo connects and stopping cleanly before it disconnects.
- Removed the dead in-memory `_worker_loop`/`_queue` compatibility layer, it was never reachable.

### Testing
- New `tests/test_task_queue_consumer.py`: dispatch, retry backoff, empty-queue polling, start/stop lifecycle, both handlers.
- Updated `test_background_worker.py` / `test_queue_migration_integration.py` for the new `enqueue_task` signature.
- `33 passed, 2 skipped` locally (skips are pre-existing, unrelated). No regressions in the rest of the suite.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -8,7 +8,7 @@ from fastapi.middleware.cors import CORSMiddleware
 
 from app.config import settings
 from app.core.logging_config import setup_logging
-from app.workers.background_worker import start_worker
+from app.workers.background_worker import start_worker, stop_worker
 from app.services.reminder_service import check_and_fire_reminders
 from app.services.digest import run_weekly_digests
 from app.services.database import get_db
@@ -160,6 +160,7 @@ async def lifespan(app: FastAPI):
 
     # Shutdown
     _scheduler.shutdown(wait=False)
+    await stop_worker()
     await close_mongo_connection()
     logger.info("Verath shut down cleanly")
 
--- a/backend/app/workers/background_worker.py
+++ b/backend/app/workers/background_worker.py
@@ -1,13 +1,12 @@
 import asyncio
 import logging
+import traceback
 import uuid
-from datetime import datetime
-from enum import Enum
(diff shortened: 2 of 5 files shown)


Chetan0e (approved):
Thanks for the contribution! This addresses the root cause by making the persistent queue actually executable instead of write-only. The consumer loop, serializable task payloads, lifecycle integration, and reuse of the existing retry/dead-letter logic make this a solid architectural improvement. The added test coverage is also appreciated. Looks good to me—approved.

PR #262, opened 2026-07-23
Fix: Registration never authenticates the user, leaving them stuck on login screen

### LINKED ISSUE #244: [BUG]: onRegisterSuccess never authenticates the user thus successful registration leaves user stuck on login screen
### Description
In `App.js`, when a user completes registration, `RegisterScreen`'s `onRegisterSuccess` prop only resets the toggle state:

```js
<RegisterScreen
  onRegisterSuccess={() => setIsRegistering(false)}
  ...
/>
```

This flips `isRegistering` back to `false`, which sends the user back to `LoginScreen`  but `isAuthenticated` is never set to `true`. So a user who just successfully registered lands back on the login form with no confirmation their account was created, and has to manually type their credentials in again to actually get into the app.

### Expected Behavior
Either:
- Registration should auto-login the user (`setIsAuthenticated(true)` on success, mirroring `onLoginSuccess`), **or**
- If manual login-after-register is intentional, the user should see explicit feedback ("Account created  please log in") rather than silently landing back on a blank login form.

### Steps to Reproduce
1. Open the app, tap "Sign up" / switch to `RegisterScreen`
2. Complete registration successfully
3. Observe: app returns to `LoginScreen` with no indication registration succeeded

### Suggested Fix
```js
<RegisterScreen
  onRegisterSuccess={() => {
    setIsRegistering(false);
    setIsAuthenticated(true); // or show a success toast if manual login is intended
  }}
  onSwitchToLogin={() => setIsRegistering(false)}
/>
```

Tech4Aditya:
Closes #244

### Problem

In `App.js`, `RegisterScreen`'s `onRegisterSuccess` only reset the screen toggle:

```js
onRegisterSuccess={() => setIsRegistering(false)}
```

This flipped `isRegistering` back to `false`, sending the user to `LoginScreen`, but `isAuthenticated` was never set to `true`. Meanwhile, in `RegisterScreen.js`, `handleRegister` only called `/auth/signup` which creates the account but returns no token — then called `onRegisterSuccess()` directly. No token was ever stored in `AsyncStorage`.

Net effect: a user who successfully registered landed back on a blank login form and had to manually re-enter their credentials to get into the app.

### Fix

**`RegisterScreen.js`**
After signup succeeds, immediately call `/auth/login` with the same credentials, store the returned `access_token` in `AsyncStorage` under `verath_token` (same key `App.js` reads on launch), and only then call `onRegisterSuccess()`. If the login step fails for any reason, fall back to the login screen with a clear "Account Created  please log in" message instead of pretending the user is authenticated.

**`App.js`**
`onRegisterSuccess` now also calls `setIsAuthenticated(true)`, mirroring `onLoginSuccess`. This is only safe because `RegisterScreen` guarantees a valid token is already in storage before this fires.

### Why not just flip `isAuthenticated` in `App.js` alone

That was my first pass and it's wrong: `isAuthenticated` would be `true` with no token in `AsyncStorage`. On next app restart, `checkAuth()` would find no `verath_token` and bounce the user straight back to login anyway. The token has to actually exist before the state flips.

### Scope

- `RegisterScreen.js`: added post-signup login call, token storage, fallback handling
- `App.js`: `onRegisterSuccess` now also authenticates, not just toggles the screen

### Testing

- Verified successful registration goes straight into the authenticated app with no manual re-login
- Verified `verath_token` is present in `AsyncStorage` after registration
- Verified app restart post-registration stays authenticated (token persists)
- Verified fallback path (login-after-signup failing) shows the login screen with a clear message instead of a silent stuck state

### DIFF
--- a/mobile/App.js
+++ b/mobile/App.js
@@ -85,7 +85,15 @@ export default function App() {
       <View style={styles.container}>
         {isRegistering ? (
           <RegisterScreen
-            onRegisterSuccess={() => setIsRegistering(false)}
+            onRegisterSuccess={() => {
+              // Mirror onLoginSuccess: registration stores the auth token
+              // the same way login does, so we just need to flip
+              // isAuthenticated here too. Previously this only reset
+              // isRegistering, which sent the user back to a blank
+              // LoginScreen instead of into the app (#244).
+              setIsRegistering(false);
+              setIsAuthenticated(true);
+            }}
             onSwitchToLogin={() => setIsRegistering(false)}
           />
         ) : (
--- a/mobile/screens/RegisterScreen.js
+++ b/mobile/screens/RegisterScreen.js
@@ -7,6 +7,7 @@ import {
 import { LinearGradient } from "expo-linear-gradient";
 import { MaterialCommunityIcons } from "@expo/vector-icons";
 import axios from "axios";
+import AsyncStorage from "@react-native-async-storage/async-storage";
 import { API_BASE_URL } from "../config";
 
 const { width } = Dimensions.get("window");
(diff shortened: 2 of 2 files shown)


Chetan0e (approved):
Looks good. Registration now establishes a real authenticated session by logging in and storing the access token before updating the authenticated state. The fallback path also handles post-registration login failure cleanly. Approved.

Issue #248, opened 2026-07-19
[Bug] Extraction pipeline: 3 independent regex/parsing bugs (correction detection, intent classification, date parsing)


payalrvs3:
## Description

Found three separate bugs in the extraction pipeline while investigating unrelated test failures. Grouping them here since they share a file, but each is fully independent - no shared logic, no ordering dependency - so I'm planning one PR per task rather than one big one.

- [ ] Correction detection misses corrections without sentence-ending punctuation 
- [ ] Intent classification: overlapping keyword patterns cause wrong intent 
- [ ] "next <weekday>" temporal expressions aren't resolved despite being an explicitly handled pattern 

Location: `backend/app/pipeline/extraction_pipeline.py`

Labels: gssoc, type:bug

payalrvs3:
Breaking each one down:

**1. Correction detection** (`_detect_correction`)
Segments are split only on sentence-ending punctuation, then a correction marker is matched only if it's anchored at the very start of a segment. Unpunctuated text ("let's meet tomorrow no no day after tomorrow") is a single segment, so a marker appearing mid-string never matches. Fix: also split immediately before a correction marker, not just after punctuation, so the marker gets its own segment boundary either way.
Fails: `test_real_world.py::test_correction_detection` (2 of 4 cases), `test_full_pipeline_speech_correction`

**2. Intent classification** (`_detect_intent` / `INTENT_PATTERNS`)
`_detect_intent` returns the first matching intent in dict order. `'meeting'` includes `\bcall\b`, so "remind me to call Sarah" matches meeting before it ever reaches `'reminder'`. Separately, `\bneed to\b` is listed under both `'deadline'` and `'task'` - deadline comes first, so task can never win that phrase. Fix: move reminder/task ahead of meeting/deadline in the dict (their markers are more specific), and delete the duplicate `\bneed to\b` from deadline.
Fails: `test_real_world.py::test_intent_classification` (2 of 5 cases)

**3. Temporal parsing** (`_parse_temporal_entities`)
`time_patterns` includes `\bnext \w+day\b` specifically to catch "next Monday" - the regex matches fine, but `dateparser.parse()` itself returns `None` for "next <weekday>" phrasing (confirmed: "Monday" alone and "next week" both parse correctly, "next Monday" doesn't, across several settings combinations). Fix: fall back to computing the date directly when dateparser returns None for a "next <weekday>" match, rather than relying on the library for this one phrasing.
[...]

PR #279, opened 2026-07-28
Fix: correction detection for unpunctuated speech

payalrvs3:
## Summary
`_detect_correction` split text into segments only on sentence-ending punctuation (`.`, `!`, `?`, `...`), then matched a correction marker only at the start of a segment. Text without punctuation (e.g. "let's meet tomorrow no no day after tomorrow") stayed a single segment, so a marker appearing mid-string was never matched.

The segment split now also breaks immediately before a correction marker, so it gets its own segment either way. The marker list is defined once and shared between the split and the match so they can't drift apart again.

Part of #248 

## Test plan
- [x] `pytest tests/test_real_world.py::test_correction_detection -v` - 4/4 pass (was 2/4)
- [x] `pytest tests/test_real_world.py::test_full_pipeline_speech_correction -v` - passes
- [x] `pytest tests/test_real_world.py -v` - nothing else regresses

### DIFF
--- a/backend/app/pipeline/extraction_pipeline.py
+++ b/backend/app/pipeline/extraction_pipeline.py
@@ -115,15 +115,20 @@ class ExtractionPipeline:
     
     def _detect_correction(self, text: str) -> tuple[bool, str]:
         """Detect speech corrections and handle them using segment analysis."""
-        # Split into segments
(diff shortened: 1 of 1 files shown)


Chetan0e (approved):
Thanks for the contribution! The issue is well identified and the fix is clean and focused. this looks good to merge, approved

PR #281, opened 2026-07-28
Fix: temporal parsing for 'next <weekday>' phrases

payalrvs3:
## Summary
`time_patterns` includes `\bnext \w+day\b` to catch phrases like "next Monday". The regex matches fine, but `dateparser.parse()` itself returns `None` for "next <weekday>" phrasing - confirmed "Monday" alone and "next week" both parse, "next <weekday>" doesn't, across all seven days.

Added a fallback that computes the date directly when dateparser returns None for one of these matches, instead of relying on the library for this one phrasing.

Part of #248.

## Test plan
- [x] `pytest "tests/test_real_world.py::test_temporal_parsing[next Monday-True]" -v` - passes
- [x] `pytest tests/test_real_world.py::test_temporal_parsing -v` - full param set, no regressions
- [x] Verified the fallback against all 7 weekdays, not just Monday

### DIFF
--- a/backend/app/pipeline/extraction_pipeline.py
+++ b/backend/app/pipeline/extraction_pipeline.py
@@ -183,6 +183,9 @@ class ExtractionPipeline:
                     'RELATIVE_BASE': datetime.utcnow(),
                     'PREFER_DATES_FROM': 'future'
                 })
+
+                if not parsed:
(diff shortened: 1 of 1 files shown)


Chetan0e (approved):
Approved

PR #298, opened 2026-08-09
fix: cache key now includes all function arguments, not just user_id/date (closes #282)

### LINKED ISSUE #282: Bug: Cache key ignores function arguments, causing cross-request data leaks
### Description
The `cached` decorator in the caching module only includes `user_id` and `date` when building the cache key. Any other function argument — positional args, or kwargs like `post_id`, `page`, `filters`, `query`, etc. — is silently ignored when generating the cache key.

This means two calls to the same decorated function with **different arguments** (but the same `user_id`) can collide on the same cache key, causing one request to receive another request's cached result.

### Steps to Reproduce
```python
@cached(ttl_seconds=300)
async def get_post(post_id: int, user_id: int):
    return await db.fetch_post(post_id, user_id)

# First call - caches result for post_id=1
result1 = await get_post(post_id=1, user_id=42)

# Second call - different post_id, same user_id
result2 = await get_post(post_id=2, user_id=42)

assert result1 != result2  # FAILS - result2 is actually result1's cached data
```

Both calls generate the cache key `"get_post:42"` because `post_id` is never included in `key_parts`. The second call incorrectly returns the first call's cached result.

### Expected Behavior
Each unique combination of function arguments should produce a distinct cache key, so different inputs never share a cached result.

### Actual Behavior
Only `user_id` and `date` are used to build the cache key. Any other argument is dropped, causing calls with different arguments (but same `user_id`/`date`) to collide and return stale/incorrect cached data.

### Root Cause
```python
key_parts = [key_prefix, func.__name__]

if 'user_id' in kwargs:
    key_parts.append(str(kwargs['user_id']))
elif args and hasattr(args[0], 'user_id'):
    key_parts.append(str(args[0].user_id))

if 'date' in kwargs:
    key_parts.append(str(kwargs['date']))

key = ":".join(key_parts)
```
(present in both `async_wrapper` and `sync_wrapper`)

### Suggested Fix
Include all arguments in the key generation instead of a hand-picked subset:
```python
key_parts = [key_prefix, func.__name__, repr(args), repr(sorted(kwargs.items()))]
key = ":".join(key_parts)
```
If certain arguments (e.g. a `db` session object) shouldn't be part of the cache key, consider adding an explicit `key_func` parameter to `cached()` so callers can opt in to custom key logic rather than having arguments silently dropped.

### Severity
High — this is a correctness/data-leak bug, not just a performance issue. It can cause one user's or one resource's cached response to be served for a different request.

Tech4Aditya:
## Description
Fixes a cache key correctness bug where the `cached` decorator only used
`user_id` and `date` when building cache keys, silently ignoring every
other argument (`post_id`, `page`, `filters`, `query`, etc.). This caused
calls with different arguments but the same `user_id` to collide on the
same cache key, resulting in cross-request data leaks — one request could
receive another request's cached result.

Closes #282

## Root cause
```python
key_parts = [key_prefix, func.__name__]
if 'user_id' in kwargs:
    key_parts.append(str(kwargs['user_id']))
elif args and hasattr(args[0], 'user_id'):
    key_parts.append(str(args[0].user_id))
if 'date' in kwargs:
    key_parts.append(str(kwargs['date']))
```
Any argument outside this hand-picked subset was dropped from the key
entirely, present in both `async_wrapper` and `sync_wrapper`.

## Fix
Replaced the hand-picked key construction with `_default_key_func`, which:
- Uses `inspect.signature(func).bind()` to bind all positional and keyword
  arguments to their parameter names, then builds the key from the full
  sorted argument set — not a subset.
- Ensures equivalent calls (e.g. `get_post(1, 42)` vs
  `get_post(post_id=1, user_id=42)`) produce the **same** key (correct —
  should share a cache entry), while calls with genuinely different
  argument values always produce **different** keys (the actual bug).
- Falls back gracefully if signature binding fails for edge-case function
  signatures (e.g. bare `*args`/`**kwargs` functions).

Also added an optional `key_func` parameter to `cached()`, so callers can
explicitly opt into excluding specific arguments (e.g. a non-hashable `db`
session object) from the cache key, rather than having them silently and
implicitly dropped as before.

## Testing
Added regression tests covering:
- Different positional arguments produce different cache entries
  (the exact repro case from the issue)
- Different keyword arguments produce different cache entries
- Identical repeated calls correctly hit the cache (no unnecessary
  re-execution)
- Positional and keyword-equivalent calls share the same cache entry
- Sync function variant behaves identically to async
- `key_func` override correctly excludes a non-hashable argument (e.g.
  a fake db session) while still varying on the arguments that matter

All existing tests in the suite pass unchanged.

## Behavioral note for reviewers
This change alters the cache key format, which means **all existing cache
entries become unreachable on deploy** — not stale/wrong, just orphaned
(they'll be pruned naturally by TTL expiry or overwritten). This is safe:
every "miss" just falls through to a fresh function call, same as a cold
cache. No stale/incorrect data can be served during the transition.

## Checklist
- [x] Bug reproduced and root cause confirmed against original code
- [x] Fix implemented in both `async_wrapper` and `sync_wrapper`
- [x] Regression tests added per @Chetan0e's request in the issue
- [x] `key_func` escape hatch added per suggested fix in the issue

### DIFF
--- a/backend/app/core/cache.py
+++ b/backend/app/core/cache.py
@@ -1,5 +1,6 @@
 import time
 import hashlib
+import inspect
 from functools import wraps
 from typing import Callable, Any, Optional
 from fastapi import Response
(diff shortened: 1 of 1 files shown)


Chetan0e (approved):
Great fix for #282. The new argument-aware cache key prevents collisions between different requests while preserving equivalent positional/keyword calls. The added key_func also provides a clean escape hatch for arguments that should not participate in caching. Good regression coverage and focused scope.

PR #313, opened 2026-08-14
fix: renew task heartbeat during handler execution

### LINKED ISSUE #296: [Bug] Task heartbeat is never renewed, long-running tasks get silently duplicate-processed
## Description
`task_queue.py`'s `dequeue()` reclaims any task stuck in `PROCESSING` whose `heartbeat_at` is older than 5 minutes (`HEARTBEAT_TIMEOUT`), resetting it to `QUEUED` so another worker can pick it up, this exists to recover tasks orphaned by a crashed worker. `update_heartbeat(task_id)` exists specifically to keep that from firing on a task that's still legitimately running. It's never called anywhere outside its own definition confirmed by `grep -rn "update_heartbeat"` across the whole backend, including `tests/`.

`background_worker.py`'s consumer loop just awaits the handler straight through:
```python
await handler(task.payload, task.user_id)
await task_queue.update_status(task.task_id, TaskStatus.COMPLETED)
```
and handlers aren't even passed `task_id`, so they structurally can't renew it themselves. Recordings are allowed up to 3600 seconds (`validate_duration`'s max), and compression tasks loop over every one of a user's memories with per-item DB writes either can easily exceed 5 minutes during completely normal operation, not just on error.

**Impact:** with more than one worker process (horizontal scaling, or a restart overlapping the old process still finishing up), a task that's genuinely still running gets reset to `QUEUED` past the 5-minute mark, gets re-claimed and re-executed by a second worker, and both copies run concurrently duplicate recordings, duplicate LLM calls, a compression job running twice. Whichever finishes second has its `update_status()` silently no-op (the `worker_id` filter no longer matches its now-stale claim), so the duplicate leaves no trace beyond a debug-level "0 rows modified." Also untested `heartbeat` only appears in `test_health.py`, in the unrelated `/status` endpoint context.

### Suggested fix: 
renew the heartbeat periodically while a handler runs, a background `asyncio.Task` pinging every ~60s alongside `await handler(...)`, cancelled once it returns or pass `task_id` into handlers so the compression loop can renew it between memories.

**Location:** `backend/app/workers/task_queue.py` (`dequeue`, `update_heartbeat`), `backend/app/workers/background_worker.py` (`_process_task`)

payalrvs3:
**Fixes #296**

#### Problem
`dequeue()` resets any task stuck in `PROCESSING` whose `heartbeat_at` is older than `HEARTBEAT_TIMEOUT` (5 minutes) back to `QUEUED`, so another worker can pick it up meant to recover tasks orphaned by a crashed worker. `update_heartbeat()` exists to keep this from firing on a task that's still legitimately running, but it was never called anywhere. The consumer loop just awaits the handler straight through, and handlers aren't even passed `task_id`, so they couldn't renew it themselves either.

Recordings are allowed up to 3600 seconds and compression loops over every one of a user's memories either can easily exceed 5 minutes in normal operation, not just on error. With more than one worker process, a task still genuinely running gets reclaimed and re-run by a second worker while the original is still going duplicate recordings, duplicate LLM calls, a compression job running twice. The losing worker's completion write silently no-ops, so this leaves no trace.

#### Fix
Run a heartbeat-renewal loop alongside the handler, cancelled once it finishes - same cancel/await/suppress pattern this file already uses in `stop_worker()`.

#### Changes
- `backend/app/workers/background_worker.py` added `HEARTBEAT_INTERVAL_SECONDS`, `_heartbeat_loop()`, wired into `_process_task()`.
- `backend/tests/test_task_queue_consumer.py` added `TestHeartbeatRenewal`: renewal during a running handler, cleanup on success and on failure.

#### Testing
- Reproduced against the original code: `update_heartbeat` called 0 times during a slow handler.
- Verified the fix: renewed periodically, cancelled cleanly on both success and failure.
- Full regression pass across every file touching these modules all green except one pre-existing, unrelated failure (confirmed to fail identically on the original code).
```
pytest tests/test_task_queue_consumer.py tests/test_background_worker.py tests/test_dead_letter_retry.py tests/test_queue_migration_integration.py tests/test_task_queue_init_race.py -v
```

### DIFF
--- a/backend/app/workers/background_worker.py
+++ b/backend/app/workers/background_worker.py
@@ -33,6 +33,7 @@ logger = logging.getLogger(__name__)
 
 POLL_INTERVAL_SECONDS = 5
 BASE_RETRY_DELAY_SECONDS = 30
+HEARTBEAT_INTERVAL_SECONDS = 60  # must stay well under TaskQueue.HEARTBEAT_TIMEOUT (5 min)
 
 _consumer_task: Optional[asyncio.Task] = None
 
@@ -117,6 +118,18 @@ TASK_HANDLERS: Dict[TaskType, Callable[[Dict[str, Any], str], Any]] = {
 
 
 # ── Consumer loop ─────────────────────────────────────────────────────────────
+async def _heartbeat_loop(task_id: str) -> None:
+    """Renew a claimed task's heartbeat every HEARTBEAT_INTERVAL_SECONDS.
+
+    Runs alongside a handler for as long as _process_task lets it; without
+    this, a handler running past TaskQueue.HEARTBEAT_TIMEOUT looks stale and
+    dequeue() reclaims it for another worker while it's still in progress.
+    """
+    while True:
+        await asyncio.sleep(HEARTBEAT_INTERVAL_SECONDS)
+        await task_queue.update_heartbeat(task_id)
+
+
 async def _process_task(task: Task) -> None:
     """Run a single claimed task and report the outcome back to task_queue."""
 
@@ -133,6 +146,7 @@ async def _process_task(task: Task) -> None:
         )
         return
 
+    heartbeat_task = asyncio.create_task(_heartbeat_loop(task.task_id))
     try:
         await handler(task.payload, task.user_id)
         await task_queue.update_status(task.task_id, TaskStatus.COMPLETED)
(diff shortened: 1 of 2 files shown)


Chetan0e (approved):
This is an important correctness/reliability fix: without heartbeat renewal, legitimate long-running tasks can be reclaimed after 5 minutes and executed by another worker. The implementation also adds cleanup and regression tests for both success and failure paths.

The scope is focused and the reported tests pass aside from a documented unrelated pre-existing failure.

PR #314, opened 2026-08-14
fix: enforce_lifecycle_limits reads importance from metadata.importance

### LINKED ISSUE #297: [Bug] `enforce_lifecycle_limits` reads importance from the wrong field - always archives instead of promoting
## Description
Memory documents store importance nested at `metadata.importance` confirmed directly in `store_memory()`, where the whole input `metadata` dict (importance included) is stored verbatim under `doc["metadata"]`; there's no top-level `importance` key anywhere in the schema.

`auto_promote_important_memories()` gets this right (`metadata.importance` in both its query and its read). Its sibling, `enforce_lifecycle_limits()` called right after it, on the same compression task, doesn't:
```python
if mem.get("importance", 0) >= self.IMPORTANCE_THRESHOLD:
    await self.promote_to_long_term(user_id, mem["_id"])
else:
    await self.archive_memory(user_id, mem["_id"])
```
`mem.get("importance", 0)` always falls through to `0`, since importance only lives at `metadata.importance`. So this always takes the `else` -  **every** excess short-term memory gets archived once a user crosses `MAX_SHORT_TERM` (100), regardless of true importance. A memory scored 0.95 is archived exactly like one scored 0.05.

This is the exact bug class `tests/test_lifecycle_field_divergence.py` exists to catch, it explicitly tests and fixes this same top-level-vs-`metadata.importance` divergence in four other methods, but `enforce_lifecycle_limits` isn't among them. It's also masked by its own test: `test_lifecycle_schema.py::test_inserting_over_100_short_term_triggers_archival` seeds fixtures with `metadata: {"importance": 0.1}` intending to hit the archive path, then only asserts `update_one.call_count == 5` never checking *which* branch fired, and never adding a high-importance fixture. Since both the correct read (0.1) and the buggy read (falls back to 0) land below the 0.6 threshold, the bug produces an identical test result and stays invisible.

### Suggested fix:
`mem.get("importance", 0)` → `mem.get("metadata", {}).get("importance", 0)`, plus a high-importance fixture in the existing test asserting promotion, not archival.

**Location:** `backend/app/db/memory_lifecycle.py` (`enforce_lifecycle_limits`), `backend/tests/test_lifecycle_schema.py`
**Suggested labels:** `gssoc`, `type:bug`,

payalrvs3:
**Fixes #297**

#### Problem
Memory documents store importance nested at `metadata.importance` confirmed in `store_memory()`, where the whole input `metadata` dict is stored verbatim under `doc["metadata"]`. There's no top-level `importance` key in the schema.

`auto_promote_important_memories()` reads this correctly. Its sibling `enforce_lifecycle_limits()` called right after it, on the same compression task reads `mem.get("importance", 0)`, which always falls back to `0`. So every excess short-term memory gets archived once a user crosses `MAX_SHORT_TERM` (100), regardless of true importance a memory scored 0.95 is archived exactly like one scored 0.05.

Same field-divergence bug `test_lifecycle_field_divergence.py` already tests and fixes in four other methods — this one wasn't among them, and its own existing test only checked that *an* update happened, not which branch fired, so it never caught this.

#### Fix
Read `metadata.importance` instead of the nonexistent top-level field - a one-line change.

#### Changes
- `backend/app/db/memory_lifecycle.py` - `enforce_lifecycle_limits()` now reads `mem.get("metadata", {}).get("importance", 0)`.
- `backend/tests/test_lifecycle_field_divergence.py` - added `TestEnforceLifecycleLimitsUsesMetadataImportance`, matching the existing pattern used for `auto_promote_important_memories`.

#### Testing
- Confirmed the new test fails against the original code (`'archived' != 'archived'` - a 0.9-importance memory got archived).
- Verified the fix: both new tests pass.
- Full regression pass on `test_lifecycle_field_divergence.py` (11/11 relevant) and `test_lifecycle_schema.py` — no existing test broken.
```
pytest tests/test_lifecycle_field_divergence.py tests/test_lifecycle_schema.py -v
```

### DIFF
--- a/backend/app/db/memory_lifecycle.py
+++ b/backend/app/db/memory_lifecycle.py
@@ -113,7 +113,7 @@ class MemoryLifecycleManager:
                 }).sort("timestamp", 1).limit(excess)
                 
                 async for mem in cursor:
-                    if mem.get("importance", 0) >= self.IMPORTANCE_THRESHOLD:
+                    if mem.get("metadata", {}).get("importance", 0) >= self.IMPORTANCE_THRESHOLD:
                         await self.promote_to_long_term(user_id, mem["_id"])
                     else:
                         await self.archive_memory(user_id, mem["_id"])
--- a/backend/tests/test_lifecycle_field_divergence.py
+++ b/backend/tests/test_lifecycle_field_divergence.py
@@ -9,6 +9,8 @@ Verifies that:
 5. auto_promote_important_memories increments promoted_count and does not archive
    memories that passed the importance gate.
 6. Full short_term → long_term → archived lifecycle transition succeeds end-to-end.
+7. enforce_lifecycle_limits reads metadata.importance (not top-level importance)
+   when deciding promote vs archive on short-term overflow.
 """
 import pytest
 from unittest.mock import AsyncMock, MagicMock, patch, call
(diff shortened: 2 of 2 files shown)


Chetan0e (approved):
This is a real correctness bug: enforce_lifecycle_limits() was reading the wrong field, causing important memories to potentially be archived incorrectly once the short-term limit was reached.

The fix is minimal and directly aligned with the existing data structure, and the contributor added a regression test specifically covering the previously missed case. The reported relevant tests also pass.

